PR #3, opened 2026-04-12
Concurrent search via `OnceLock` lazy caches, with x86 heap fill overflow fix

nicolas-geysse:
````md
This PR has two commits that address related findings: a latent
bug in the x86 top-k fill path, and a refactor that lets `search`
take `&self` so a single `TurboQuantIndex` can serve concurrent
queries from multiple threads.

## 1. `fix(search): x86 heap fill overflow when k < BLOCK`

The x86_64 AVX2 search path had a per-block top-k fill loop that
checked `*sz < k` once **before** the lane loop but not inside it.
Once the heap reached capacity mid-block the next iteration wrote
`hs[*sz]` at `*sz = k` and panicked with an out-of-bounds index
access.

This is hit whenever `k < BLOCK (= 32)`: the first block scores
up to 32 lanes but the heap only holds `k` entries. With `k = 10`
the 11th fill write overflows. The benchmarks in `benchmarks/suite/`
all use `k = 64` (> `BLOCK`) so the bug never surfaces there, yet
`k` between 5 and 20 is exactly the shape of queries a RAG
application issues — a user typically asks for the top 5 to 10
relevant documents.

The ARM NEON path at the same level of the file already handles
this correctly (see the inline heap in `score_*_block_neon`
callers): it checks `if heap_sizes[qi_off] < k` per lane inside
the loop and falls back to the scalar heap-update branch when
the fill phase ends mid-block. This commit ports that per-lane
check to the x86 path.

Minimal reproducer (panics on `main`, passes with this fix):

```rust
use turbovec::TurboQuantIndex;

let mut index = TurboQuantIndex::new(256, 4);
let vectors: Vec<f32> = (0..1024 * 256).map(|i| i as f32).collect();
index.add(&vectors);
let query: Vec<f32> = vec![0.0; 256];
let _ = index.search(&query, 10);
````

Diff: +23 / -5 lines in turbovec/src/search.rs.

## 2. refactor: OnceLock-based lazy caches, search(&self) for concurrent use

Before this change `TurboQuantIndex::search` took `&mut self`
because the rotation matrix, the Lloyd-Max centroids and the
SIMD-blocked code layout were stored as `Option<Vec<_>>` fields
and populated on first call via `ensure_cached`. This forced
callers to serialise every search on the same index behind a
mutex, which is fine for a single-threaded benchmark but
unworkable for a multi-user application that expects concurrent
top-k queries on a shared index.

The fix replaces the three caches with `std::sync::OnceLock`
(stable since Rust 1.70) so initialisation is lazy AND thread-safe
from `&self`:

* rotation:  `OnceLock<Vec<f32>>`
* centroids: `OnceLock<Vec<f32>>`
* blocked:   `OnceLock<BlockedCache>` (data + n_blocks bundled)

`search` now takes `&self` and multiple threads can call it on a
shared `Arc<TurboQuantIndex>` without external locking. The first
concurrent caller pays the one-time init via `OnceLock::get_or_init`
while the others block briefly, then every subsequent read is
lock-free.

`add` still takes `&mut self` because it extends `packed_codes`
and `norms`. It invalidates the blocked cache by replacing its
`OnceLock` — the invariant is that once a cache is populated from
`&self` it matches the current `packed_codes`. `rotation` and
`centroids` are deterministic functions of `(dim, ROTATION_SEED)`
and `(bit_width, dim)` respectively, so they never need to be
reset.

A new `prepare()` method forces eager cache initialisation. It is
safe to call from multiple threads and is useful right after
`load()` or after a batch of `add()` calls to pay the one-time
cost out of the first query path. Calling it is optional — the
`OnceLock`-backed lazy path is correct either way.

The Python binding gets the same treatment: `search` on the
pyclass is now `&self` and a `prepare()` method is exposed.
Existing Python callers see no behavioural change.

### Tests (`turbovec/tests/concurrent_search.rs`)

* `search_is_deterministic_across_threads` — 16 threads × 32
  searches against a shared `Arc<TurboQuantIndex>` all return the
  same top-k as the main thread. Catches visibility and
  initialisation races in the `OnceLock` caches.
* `lazy_init_is_safe_when_prepare_is_skipped` — 32 threads hit
  `get_or_init` simultaneously on a fresh index, exercising the
  one-winner-many-readers race.
* `prepare_is_idempotent_from_multiple_threads` — eager init from
  8 concurrent callers, then a follow-up search to confirm the
  index is still usable.
* `add_after_search_invalidates_blocked_cache` — after a search
  that populates the cache, an `add` must reset `blocked` so
  subsequent searches see the new vectors.
* `write_load_preserves_concurrent_search_results` — roundtrip
  through the `.tv` file format returns the same top-k for the
  same queries.

Runs in ~2.8 s on a 24-core x86_64 host with the standard
`cargo test -p turbovec` invocation (after adding
`-lopenblas` to `RUSTFLAGS` locally — same requirement as
existing upstream benchmarks).

Diff: +324 / -54 lines across `turbovec/src/lib.rs`,
`turbovec-python/src/lib.rs`, and the new integration test.

## Relation between the two commits

The concurrent-search tests use `k = 3, 5, 8 and 10` —
realistic values for a top-k vector search. These values trigger
the x86 fill overflow described in commit 1, which is why the
two commits are presented together. The fix can be cherry-picked
independently if desired.

## MSRV

`std::sync::OnceLock` is stable since Rust 1.70 (June 2023). The
crate has no explicit `rust-version` today; happy to add
`rust-version = "1.70"` to `turbovec/Cargo.toml` as a follow-up
if you want the MSRV recorded.

## Found during

Integrating turbovec into a multi-user RAG backend where a single
`TurboQuantIndex` shard is shared across concurrent HTTP handlers.
The `&mut self` on `search` made a shared-index design impossible;
the x86 fill overflow then blocked the tests we wrote to validate
the refactor. Fixing both felt cleaner than working around either.

Thanks for the library — the data-oblivious rotation + Lloyd-Max
design is a great fit for incremental indexes and we're excited
to ship it in production.

```
```

### DIFF
--- a/turbovec-python/src/lib.rs
+++ b/turbovec-python/src/lib.rs
@@ -23,7 +23,7 @@ impl TurboQuantIndex {
     }
 
     fn search<'py>(
-        &mut self,
+        &self,
         py: Python<'py>,
         queries: PyReadonlyArray2<f32>,
         k: usize,
@@ -55,6 +55,13 @@ impl TurboQuantIndex {
             pyo3::exceptions::PyIOError::new_err(format!("{}", e))
         })?;
         Ok(Self { inner })
+    }
+
+    /// Warm up the search caches (rotation matrix, Lloyd-Max centroids,
+    /// SIMD-blocked code layout) so the first `search` call does not pay
+    /// the one-time initialisation cost.
+    fn prepare(&self) {
+        self.inner.prepare();
     }
 
     fn __len__(&self) -> usize {
--- a/turbovec/src/lib.rs
+++ b/turbovec/src/lib.rs
@@ -3,15 +3,36 @@
 //! Compresses high-dimensional vectors to 2-4 bits per coordinate with
 //! near-optimal distortion. Data-oblivious — no training required.
 //!
-//! ```rust
+//! ```no_run
 //! use turbovec::TurboQuantIndex;
 //!
+//! // 1536-dim vectors compressed to 4 bits per coordinate.
 //! let mut index = TurboQuantIndex::new(1536, 4);
+//!
+//! // `vectors` is a flat [f32] of length n * dim, `queries` likewise.
+//! let vectors: Vec<f32> = vec![0.0; 1536 * 10];
+//! let queries: Vec<f32> = vec![0.0; 1536 * 2];
+//!
 //! index.add(&vectors);
 //! let results = index.search(&queries, 10);
 //! index.write("index.tv").unwrap();
 //! let loaded = TurboQuantIndex::load("index.tv").unwrap();
 //! ```
+//!
+//! # Concurrent search
+//!
+//! `search` takes `&self` and is safe to call from multiple threads
+//! concurrently. Internally the rotation matrix, the Lloyd-Max centroids
+//! and the SIMD-blocked code layout are initialised lazily via
+//! [`std::sync::OnceLock`], so the first caller pays the one-time
+//! initialisation cost and every subsequent caller reads the caches
+//! without locking. [`TurboQuantIndex::prepare`] can be called once
+//! after `add`/`load` to pay that cost up front.
+//!
+//! Mutation still flows through `&mut self`: `add` extends the packed
+//! codes and invalidates the blocked layout cache by replacing its
+//! `OnceLock`. This keeps the invariant that once a cache is populated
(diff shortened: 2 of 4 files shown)


RyanCodrai (approved):

RyanCodrai:
Great PR thank you

PR #6, opened 2026-04-17
feat: Add dynamic delete support

Pringled:
Thanks for making this great library! While trying it out, I noticed that dynamic deletion isn't supported, but it's trivial to support that with this algorithm, hence this PR. Since very few ANN backends support dynamic deletion I think this adds a lot of value to this library.

The way it's set up is as follows: the target slot is overwritten with the last vector's packed codes and norm, both arrays are truncated by one, and the blocked layout cache is invalidated so the next search rebuilds against the updated data. This is an O(1) operation since it's swap-and-pop deletion.

Note: I didn't add tests since I didn't see tests in the existing package.

### DIFF
--- a/turbovec-python/src/lib.rs
+++ b/turbovec-python/src/lib.rs
@@ -64,6 +64,14 @@ impl TurboQuantIndex {
         self.inner.prepare();
     }
 
+    /// Remove the vector at `idx` (swap-and-pop).
+    ///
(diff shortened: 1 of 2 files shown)


RyanCodrai:
@Pringled 
Thank you for your PR! Great contribution suggestion. I created a branch off your contribution so that you got listed as a contributor :) thanks!

PR #22, opened 2026-05-16
Search-time filtering: mask= / allowlist= kwargs (closes #21)

### LINKED ISSUE #21: Feature request: Pre-filtering for hybrid search
### Description
Adding an optional pre-filtering parameter to the `search` method would be incredibly helpful for hybrid search workloads, ensuring valid candidates aren't dropped by post-filtering when specific metadata constraints are known upfront.

```python
import numpy as np

# e.g., IDs retrieved from an external SQL DB condition
allowed_ids = np.array([1001, 1003, 1008], dtype=np.uint64)
scores, ids = id_map_index.search(query, k=10, allowlist=allowed_ids)

# The mask length matches the number of vectors in the index.
mask = np.ones(len(tq_index), dtype=bool)
mask[[0, 2, 5]] = False  # Disable specific slots
scores, indices = tq_index.search(query, k=10, bitmask=mask)


```

Thanks for the great work on this library!

RyanCodrai:
## Summary

Adds optional `mask=` / `allowlist=` kwargs to `search` on both index
types so callers can restrict the returned top-k to a caller-supplied
subset of vectors. The kernel applies the filter at the heap-update
site rather than via post-filtering, so selective filters return up to
`k` results from the allowed set instead of fewer-than-`k` from an
over-fetch pass.

Closes #21.

## What's in

**Core feature** (`feb645c`)
- `TurboQuantIndex::search_with_mask(queries, k, Option<&[bool]>)` — slot bitmask.
- `IdMapIndex::search_with_allowlist(queries, k, Option<&[u64]>)` — id allowlist; builds the bitmask via `id_to_slot`.
- Python: keyword-only `mask=` / `allowlist=` on the corresponding `search` methods. Pre-validates dtype, shape, emptiness and unknown ids to raise clean `ValueError` / `KeyError` rather than panic.
- Threaded through every scoring path: NEON (aarch64), AVX2 (x86_64), AVX-512BW (x86_64), and the scalar fallback. NEON's check sits in the per-query top-k scan; the x86 paths gate each inline heap update.
- `n_allowed == 0` short-circuits before the kernel so empty-heap update sites stay safe.
- Output shape shrinks to `min(k, n_allowed)` — consistent with the existing `k > len(idx)` contract. No sentinel padding.

**Docs + integrations** (`cf009c0`)
- `docs/api.md`: kwargs documented + new Filtering section.
- `README.md`: feature bullet + hybrid-retrieval example.
- `CHANGELOG.md`: new file under [Unreleased].
- Haystack `embedding_retrieval`: drops the `top_k * 10` over-fetch hack; resolves filters to an allowlist upfront. The "results may be fewer than top_k" caveat in the docstring goes away.
- LangChain `similarity_search*`: accepts `filter: dict | Callable | None` (previously silently ignored via `**_`).
- LlamaIndex `query()`: honours `VectorStoreQuery.filters` (EQ, NE, GT/GTE/LT/LTE, IN/NIN, TEXT_MATCH, CONTAINS, IS_EMPTY; AND/OR; nested) and `VectorStoreQuery.doc_ids`. Unsupported operators raise `NotImplementedError` rather than silently mismatching.

## Tests

- Rust: 11 new tests in `turbovec/tests/filtering.rs` covering correctness vs post-hoc reference, `mask=None` equivalence, effective-k shrinking, multi-query batches across the 4-query NEON kernel and its tail, error paths, and id stability across `swap_remove`.
- Python bindings: 14 tests in `turbovec-python/tests/test_filtering.py` (shape/dtype validation, kwarg-only enforcement, parity with reference filter).
- Integration filter tests added to `test_haystack.py` / `test_langchain.py` / `test_llama_index.py`, including a regression test that a filter matching 3 of 50 docs returns all 3 with `top_k=3`.

Full suite (locally on ARM/macOS): Rust crate + 97 Python tests pass.

## Caveats

The x86 paths (AVX2 + AVX-512BW) compile cleanly but I couldn't run them on this machine. The kernel changes are mechanical — same `if let Some(m) = mask { if !mask_allows(...) { continue; } }` guard at every heap-insert site — but they should get a smoke test on x86 before merge.

## Test plan

- [ ] `cargo test -p turbovec` on the x86 benchmark instance
- [ ] `pytest turbovec-python/tests` on the x86 benchmark instance
- [ ] Spot-check a quick speed bench to confirm `mask=None` doesn't regress unfiltered search

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,5 @@
+{
+  "worktree": {
+    "bgIsolation": "none"
+  }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ __pycache__/
 *.pyc
 *.bak
 *.so
+*.dSYM/
 *.tq
 *.npy
 codebooks.npz
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,35 @@ appears under each surface it touches.
   - Threaded through every scoring path: NEON (aarch64), AVX2
     (x86_64), AVX-512BW (x86_64), and the scalar fallback.
 
+- **Lazy index construction.** The dim can now be deferred and inferred
+  from the first batch of vectors, rather than committed at construction
+  time. This is the same ergonomic improvement integration users were
+  already getting through the framework wrappers, pulled down into the
+  core so direct Rust users and any future integration get it for free.
+  - `TurboQuantIndex::new_lazy(bit_width)` and
+    `IdMapIndex::new_lazy(bit_width)` — construct an empty index with
+    no committed dim.
+  - `TurboQuantIndex::add_2d(vectors, dim)` and
+    `IdMapIndex::add_with_ids_2d(vectors, dim, ids)` — add a flat
+    vector batch with an explicit dim; locks the index dim on the
+    first call, validates on subsequent ones. Existing `add(&[f32])` /
+    `add_with_ids(&[f32], &[u64])` still work on a dim-known index and
+    panic with a clear message on a lazy uncommitted one.
+  - `TurboQuantIndex::dim_opt()` / `IdMapIndex::dim_opt()` return
+    `Option<usize>` — `None` for the lazy uncommitted state. The
+    existing `dim() -> usize` getters keep returning `usize`, with `0`
+    as a non-breaking sentinel for the lazy state (the eager
+    constructor asserts `dim >= 8`, so `0` doesn't collide).
+  - File format: `.tv` and `.tvim` headers encode the lazy state via
+    a `dim = 0` sentinel. Files written before this change always have
+    `dim >= 8` and load cleanly into the eager state.
+
+#### Changed
+
+- `search`, `search_with_mask`, and `prepare` on `TurboQuantIndex`
+  return empty results / are no-ops when called on a lazy
+  uncommitted index, rather than panicking.
+
 ### turbovec — Python package (current: 0.3.0 → next: 0.4.0)
 
 #### Added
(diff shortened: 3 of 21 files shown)


RyanCodrai:
## Review checklist & walkthrough

You don't need to read any Rust to verify this — everything below is pitched at the conceptual level, with pointers to the Python-side files where the visible API lives. I've split it into: how the feature works, what you should personally check, every test added, integration changes, the changelog, and the README.

---

## How the filtering works (high-level)

**The problem.** Today, if you want "search for the 10 nearest vectors, but only among these 500 specific ids", you'd have to either (a) post-filter — search for many more than 10, then drop non-matches, hoping enough survive (the Haystack integration's old behaviour, with `top_k * 10` over-fetch), or (b) accept fewer than 10 results when the filter is selective. Both are bad.

**The fix.** Push the filter into the scoring loop. The kernel scores vectors in 32-wide blocks; at the end of each block it considers inserting them into a top-k heap. We add a check at that insert step: \"is this slot allowed?\" If no, skip the insert; if yes, proceed. The block scoring itself is unchanged — only the heap insertion is gated.

**Why this is correct.** A vector that's never inserted into the heap can never appear in the final top-k. So the result is exactly the top-k *within the allowed set*, with no recall loss from over-fetching too few candidates.

**The data flow, from user API to kernel.**

```
User (Python)
  │
  │  TurboQuantIndex.search(queries, k, mask=bool_array)
  │  IdMapIndex.search(queries, k, allowlist=uint64_array)
  │
  ▼
Python binding (turbovec-python/src/lib.rs)
  │  - Validates shape / dtype / emptiness
  │  - For allowlist: looks each id up to confirm it exists, builds a bool array
  │  - Passes a bool slice into the Rust core
  │
  ▼
Rust core (turbovec/src/lib.rs, id_map.rs)
  │  - Packs the bool array into a tight bitset (1 bit per slot, 64 bits per machine word)
  │  - Counts the set bits → effective_k = min(k, n_allowed)
  │  - Hands the bitset to the kernel
  │
  ▼
Scoring kernel (turbovec/src/search.rs)
  │  - At every heap-insert site, checks: is this slot's bit set in the bitset?
  │  - If yes: insert (existing logic); if no: skip
  │  - Happens uniformly across NEON (your Mac), AVX2, AVX-512BW (the GCP box), and scalar
```

**Design choices worth noting:**

1. **One bit per slot, packed into 64-bit words.** Compact and cache-friendly. Check at heap-update time is `(word >> bit) & 1` — one shift, one AND.
2. **The user-facing API differs by index type.** `TurboQuantIndex` works in raw slot indices, so it takes a `bool` array indexed by slot. `IdMapIndex` works in stable external ids, so it takes a `uint64` array of allowed ids; internally that gets translated to a bool array via the existing `id_to_slot` hash map. This mirrors the existing `remove(slot)` vs `remove(id)` split.
3. **Effective `k` shrinks instead of padding.** If you ask for `k=10` but the allowlist only has 3 ids, you get back a `(nq, 3)` array, not `(nq, 10)` with sentinel padding. This matches the existing behaviour when `k > len(index)`.
4. **Empty mask short-circuits before the kernel runs.** Defensive — keeps the heap-update sites safe.
5. **Errors are pre-validated in the Python binding layer.** Length mismatch / empty allowlist / unknown id raise `ValueError` / `KeyError` directly, not a Rust panic wrapped in `pyo3.PanicException`.

---

## What you personally should check

You can verify all of this without reading Rust:

- [ ] **The Python API reads how you want.** Skim `turbovec-python/tests/test_filtering.py` — it shows every API shape: `idx.search(queries, k, mask=...)`, `idx.search(queries, k, allowlist=...)`, error cases, edge cases.
- [ ] **Hybrid retrieval example in the README** (`README.md`, the new \"Hybrid retrieval (filtered search)\" section) — does it match how you'd want a new user to learn about this?
- [ ] **The `docs/api.md` filtering section** reads cleanly and the new kwargs in the method tables are accurate.
- [ ] **Haystack integration:** open `turbovec-python/python/turbovec/haystack.py` and look at the new `embedding_retrieval`. The old behaviour was \"over-fetch `top_k * 10`, then drop non-matches\". The new code resolves filters upfront and passes an allowlist to the kernel. The docstring no longer warns about results being fewer than `top_k` — because they won't be.
- [ ] **LangChain integration:** `turbovec-python/python/turbovec/langchain.py`. The `similarity_search` family previously silently swallowed any `filter` kwarg via `**_`. It now accepts `filter` as either a dict of metadata key/value pairs (AND semantics) or a callable predicate. Worth checking whether the dict semantics match what your LangChain users will expect — I chose \"AND of exact-equality\" which is the most common convention.
- [ ] **LlamaIndex integration:** `turbovec-python/python/turbovec/llama_index.py`. The `query()` method now honours `VectorStoreQuery.filters` and `.doc_ids`. The `_filters_match` / `_single_filter_match` helpers walk the LlamaIndex `MetadataFilters` tree. I supported what I think are the common operators (EQ, NE, GT, LT, GTE, LTE, IN, NIN, TEXT_MATCH, CONTAINS, IS_EMPTY) plus AND/OR. Anything else raises `NotImplementedError` — check whether the supported set is too narrow / too wide for your taste.
- [ ] **The CHANGELOG.md** structure (per-surface grouping, projected next versions) reads the way you want — and the wording on each bullet is right.
- [ ] **The README feature bullet** (search for \"Filter at search time\") — is the framing right?
- [ ] **The PR's commit history.** Two commits: core kernel/API/bindings, then docs+integrations+changelog. Then a small fixup for the changelog restructure. Easy to revert one without the other if needed.
- [ ] **x86 paths.** I can't run AVX2 / AVX-512 locally on the Mac. They compile cleanly and the changes are mechanical, but they want a smoke test on the GCP box before merge.

---

## Tests added (44 new tests across 5 files)

### Core feature — Rust (`turbovec/tests/filtering.rs`, 11 tests)

These prove the kernel logic is correct on this machine (ARM/NEON):

1. `mask_matches_post_hoc_filter` — masked search returns the same top-k as searching everything and filtering after. The gold-standard correctness check.
2. `mask_none_equals_mask_all_true` — `mask=None` and `mask=Some(all true)` produce identical results.
3. `effective_k_shrinks_when_allowlist_smaller_than_k` — if the mask allows 3 slots and you ask for k=10, you get 3 results.
4. `all_false_mask_returns_empty_results` — degenerate case, shouldn't panic.
5. `mask_length_mismatch_panics` — passing a mask of the wrong length is caught.
6. `multi_query_batch_respects_mask` — the kernel batches queries in groups of 4 on both NEON and x86. This test runs 7 queries (forcing both the fused 4-query path and the single-query tail) and checks every result respects the mask.
7. `allowlist_returns_only_listed_ids` — IdMapIndex variant, every returned id is in the allowlist.
8. `allowlist_none_equivalent_to_plain_search` — backward compat.
9. `empty_allowlist_panics` — validation.
10. `unknown_id_in_allowlist_panics` — validation.
11. `allowlist_survives_swap_remove` — deleting an id NOT in the allowlist doesn't invalidate the allowlist, because ids are stable across deletions even though slot positions aren't.

### Core feature — Python bindings (`turbovec-python/tests/test_filtering.py`, 14 tests)

These prove the binding layer (PyO3) does the right thing:

1. `test_mask_none_matches_unmasked` — `mask=None` behaves identically.
2. `test_mask_all_true_matches_unmasked` — all-True mask is a no-op.
3. `test_mask_restricts_returned_indices` — every returned slot is in the mask.
4. `test_mask_shrinks_effective_k` — output shape is `(nq, min(k, popcount(mask)))`.
5. `test_mask_all_false_returns_empty_columns` — shape `(nq, 0)`.
6. `test_mask_wrong_length_raises` — clean `ValueError`, not a panic.
7. `test_mask_must_be_bool_dtype` — refuses uint8 / other dtypes.
8. `test_mask_matches_post_hoc_filter` — Python-side parity check vs reference computation.
9. `test_allowlist_none_matches_unfiltered` — backward compat.
10. `test_allowlist_restricts_returned_ids` — every returned id is in the allowlist.
11. `test_allowlist_empty_raises_value_error` — clean `ValueError`.
12. `test_allowlist_unknown_id_raises_key_error` — clean `KeyError`.
13. `test_allowlist_kwarg_is_keyword_only` — can't pass positionally (catches API drift).
14. `test_mask_kwarg_is_keyword_only` — same.

### Haystack integration tests (`test_haystack.py`, 3 new)

Added to the existing 17:

1. `test_embedding_retrieval_selective_filter_returns_top_k` — the regression test for the old over-fetch hack. Filter matches 3 of 50 docs; `top_k=3`. Must return all 3. The old code could miss some when they weren't in the over-fetched window.
2. `test_embedding_retrieval_no_matches_returns_empty` — filter matching nothing returns `[]`.
3. `test_embedding_retrieval_top_k_larger_than_matches` — `top_k=100`, filter matches 10. Returns 10.

### LangChain integration tests (`test_langchain.py`, 5 new)

1. `test_similarity_search_with_dict_filter` — dict syntax, AND-of-equality on metadata.
2. `test_similarity_search_with_callable_filter` — predicate function over metadata.
3. `test_similarity_search_filter_with_scores` — filter also works in the `similarity_search_with_score` variant.
4. `test_similarity_search_filter_no_matches_returns_empty` — empty list.
5. `test_similarity_search_filter_invalid_type_raises` — passing a number (not dict / not callable) raises `TypeError`.

### LlamaIndex integration tests (`test_llama_index.py`, 8 new)

1. `test_query_with_eq_filter` — basic `MetadataFilter(operator=EQ)`.
2. `test_query_with_in_filter` — `operator=IN`, value is a list.
3. `test_query_with_and_filter` — two filters combined with `FilterCondition.AND`.
4. `test_query_with_or_filter` — two filters combined with `FilterCondition.OR`.
5. `test_query_filter_no_matches_returns_empty` — empty result.
6. `test_query_filter_selective_returns_top_k_from_matches` — the recall regression test (50 docs, 3 match, top_k=3 returns 3).
7. `test_query_with_doc_ids_filter` — `VectorStoreQuery.doc_ids` restriction (alternative to `filters`).
8. `test_query_unsupported_filter_operator_raises` — passing `FilterOperator.ANY` raises `NotImplementedError` (so silently mismatching can't happen).

---

## How each integration was wired up

### Haystack (`turbovec-python/python/turbovec/haystack.py`)

The store already stored docs in an in-memory dict keyed by an internal uint64 \"handle\" (with a side mapping to the user's string id and metadata). The old `embedding_retrieval` code:

```python
fetch_k = min(top_k * 10, count_documents())   # over-fetch hack
scores, handles = self._index.search(qvec, fetch_k)
# … iterate, drop non-matching docs, stop at top_k …
```

The new code:

```python
if filters is None:
    # No filter: just search normally.
    scores, handles = self._index.search(qvec, top_k)
else:
    # Walk in-memory doc table once, collect handles whose docs match.
    allowed_handles = [
        handle for handle, data in self._u64_to_doc.items()
        if document_matches_filter(filters, self._reconstruct(data))
    ]
    if not allowed_handles:
        return []
    allowlist = np.asarray(allowed_handles, dtype=np.uint64)
    scores, handles = self._index.search(qvec, top_k, allowlist=allowlist)
```

Same total cost (you still walk the doc table once to apply the filter), but the kernel now scores only matching docs and returns up to `top_k` results from them rather than from an over-fetched window.

### LangChain (`turbovec-python/python/turbovec/langchain.py`)

`similarity_search`, `similarity_search_with_score`, and `similarity_search_by_vector` previously had `**_: Any` to silently swallow framework kwargs. Now they accept `filter` explicitly:

```python
def similarity_search(
    self, query: str, k: int = 4,
    filter: dict[str, Any] | Callable[[dict[str, Any]], bool] | None = None,
    **_: Any,
) -> list[Document]:
    …
```

`_compile_filter` turns either form into a predicate over the metadata dict, then `_search_vector` walks `self._docs` to build the allowed-handles list and passes it through. Dict semantics are AND-of-exact-equality (`{\"tier\": \"pro\", \"region\": \"us\"}` matches docs where both keys equal those values).

### LlamaIndex (`turbovec-python/python/turbovec/llama_index.py`)

LlamaIndex's filter system is richer than LangChain's. `VectorStoreQuery` carries both `.filters` (a `MetadataFilters` tree) and `.doc_ids` (a list of node ids to keep).

I added `_resolve_allowed_handles` that:

1. Honours `query.doc_ids` first (intersects with the node table).
2. Then applies `query.filters` via `_filters_match`, which recurses through the `MetadataFilters` tree (so nested AND/OR works).
3. `_single_filter_match` handles individual `MetadataFilter` operators: `EQ`, `NE`, `GT`, `LT`, `GTE`, `LTE`, `IN`, `NIN`, `TEXT_MATCH`, `CONTAINS`, `IS_EMPTY`.
4. Returns the list of u64 handles to pass to the kernel as `allowlist=`.

Anything outside that operator list raises `NotImplementedError`. I'd rather refuse loudly than silently return wrong results.

---

## CHANGELOG structure

New file at `CHANGELOG.md`, following [Keep a Changelog](https://keepachangelog.com/) format.

The header explains that the Rust crate and the Python package version independently, and that each release section is split by surface. Each section reads:

```
## [Unreleased]

### turbovec — Rust crate (current: 0.2.0 → next: 0.3.0)
  #### Added
    - Search-time filtering (search_with_mask / search_with_allowlist)

### turbovec — Python package (current: 0.3.0 → next: 0.4.0)
  #### Added
    - Search-time filtering (mask= / allowlist= kwargs)
  #### Changed
    - Haystack integration: filters → allowlist
    - LangChain integration: filter= kwarg now honoured
    - LlamaIndex integration: VectorStoreQuery.filters now honoured
```

The search-filtering feature appears under **both** surfaces (same work, different APIs). The integration changes only appear under Python (they're Python-only). The \"next:\" annotations document the expected bump but I didn't actually change the `Cargo.toml` / `pyproject.toml` versions — past convention in this repo (commit `80ebdd9`) is to bump versions in a dedicated release commit.

---

## README changes

There's only one README — `turbovec-python/README.md` is a symlink to the root `README.md`. Two changes:

1. **New feature bullet** added between \"Faster than FAISS\" and \"Pure local\":

   > **Filter at search time.** Pass an id allowlist (or a slot bitmask) to `search()` and the kernel honours it directly. You always get up to `k` results from the allowed set — no over-fetching, no recall hit on selective filters.

2. **New \"Hybrid retrieval (filtered search)\" subsection** in the Python usage section. Shows the typical pattern (external system narrows to candidate ids, then dense rerank within those):

   ```python
   allowed = np.array(
       db.execute(\"SELECT id FROM docs WHERE tenant=?\", (t,)).fetchall(),
       dtype=np.uint64,
   )
   scores, ids = idx.search(query, k=10, allowlist=allowed)
   ```

   Plus a one-line explanation that shrinking to `len(allowed)` is the expected behaviour when the allowed set is smaller than `k`.

No other README changes — keeping the rest stable. The recall/speed charts and the \"How it works\" section are untouched.

RyanCodrai:
## Integration audit — what the cross-check turned up

Your push-back was right. I audited each integration against the installed framework source (`langchain_core 0.3.63`, `llama_index.core 0.12.39`, `haystack-ai 2.28.0`) and found real divergences. Fixed in `209a1db`.

### Haystack — clean for this PR

`document_matches_filter(filters, doc)` argument order, `Document.meta` attribute name, and `embedding_retrieval` signature all check out against `InMemoryDocumentStore` (`document_stores/in_memory/document_store.py:712`). The pre-existing `scale_score` formula in `_reconstruct` is wrong (it does `sigmoid(score)` instead of `(score + 1) / 2` for cosine or `sigmoid(score / 100)` for dot-product, per `_compute_query_embedding_similarity_scores` at line 818) — but that code predates this PR.

### LangChain — one real bug

**Callable filter signature was wrong.** I implemented `Callable[[dict], bool]` (predicate over metadata dict). The in-tree `InMemoryVectorStore` uses `Callable[[Document], bool]` (predicate over the Document object) — see `langchain_core/vectorstores/in_memory.py:493`. A user porting code from the reference store would have hit `AttributeError` on `.metadata`.

Fixed by changing the callable type to `Callable[[Document], bool]`, constructing a `Document` per candidate inside the filter check, and updating tests accordingly. Added one new test that exercises `doc.page_content` access in the predicate to verify the full Document is reachable.

### LlamaIndex — four real bugs

Audited against `SimpleVectorStore._build_metadata_filter_fn` (`llama_index/core/vector_stores/simple.py:46`) and `SimpleVectorStore.query` (line 320).

1. **`VectorStoreQuery.node_ids` was ignored.** The reference store uses it as the canonical "restrict to specific nodes" knob. Now honoured.
2. **`VectorStoreQuery.doc_ids` conflated node_id and ref_doc_id.** My code matched against both. Per the LlamaIndex data model, `doc_ids` is for source-document ids (`ref_doc_id`); `node_ids` is for chunk ids. Now `doc_ids` filters by `ref_doc_id` only.
3. **`NE` returned `True` for missing keys.** The reference returns `False` — every operator except `IS_EMPTY` treats `metadata.get(key) is None` as "no match". My behaviour was the intuitive one but disagreed with the spec. Fixed.
4. **`TEXT_MATCH` was case-sensitive.** Reference is case-insensitive: `value.lower() in metadata_value.lower()`. Fixed.

Nested `MetadataFilters` are still supported (the reference rejects them with `ValueError`); that's a strict superset and useful enough to keep.

The `NOT` filter condition is still unsupported — I'd rather raise loudly than guess at semantics.

### New tests added by the audit

- `test_similarity_search_callable_filter_can_use_page_content` (LangChain) — proves the full Document is reachable in the predicate.
- `test_query_with_node_ids_filter` (LlamaIndex) — replaces the old, semantically-incorrect `test_query_with_doc_ids_filter`.
- `test_query_with_doc_ids_filter_matches_ref_doc_id_only` — verifies `doc_ids` doesn't fall through to node_id matching.
- `test_query_with_node_ids_and_filters_intersect` — combining `node_ids` with metadata filters intersects rather than unions.
- `test_query_ne_filter_treats_missing_key_as_no_match` — locks in reference-matching `NE` semantics.
- `test_query_text_match_is_case_insensitive` — locks in case-insensitive substring match.

### Test status

- Rust: full suite green.
- Python: 102 tests, all passing.

### CHANGELOG updated

The Python entries for the LangChain and LlamaIndex integrations were rewritten to describe the actual (now-corrected) semantics: callable filter is over `Document`, `node_ids` is honoured, doc_ids matches `ref_doc_id` only, missing-key handling matches `SimpleVectorStore`, `TEXT_MATCH` is case-insensitive, and unsupported operators / conditions raise loudly.

PR #24, opened 2026-05-17
Add Agno framework integration

### LINKED ISSUE #23: Add Agno framework integration
Add turbovec integration for the Agno agent framework, following the same pattern as the existing LangChain, LlamaIndex, and Haystack integrations.

Agno (https://github.com/agno-agi/agno) is an agent framework with a VectorDb interface for knowledge retrieval. This integration implements that interface backed by IdMapIndex.

Proposed changes:
- turbovec-python/python/turbovec/agno.py
- turbovec-python/tests/test_agno.py
- docs/integrations/agno.md
- pyproject.toml: agno optional dependency
- README.md: add Agno to framework list

Install: pip install turbovec[agno]

raghavenderreddygrudhanti:
Fixes #23

Add turbovec integration for the Agno agent framework (https://github.com/agno-agi/agno), implementing the VectorDb interface backed by IdMapIndex.

Follows the same pattern as existing LangChain/LlamaIndex/Haystack integrations.

## Changes

- turbovec-python/python/turbovec/agno.py - TurboQuantVectorDb class
- turbovec-python/tests/test_agno.py - 14 unit tests
- docs/integrations/agno.md - usage documentation
- turbovec-python/pyproject.toml - added agno optional dependency
- README.md - added Agno to framework integrations list

## Usage

    pip install turbovec[agno]

    from turbovec.agno import TurboQuantVectorDb

    vector_db = TurboQuantVectorDb(dim=1536, bit_width=4, embedder=my_embedder)
    vector_db.create()

## Operations supported

- insert / upsert with content hash deduplication
- similarity search via embedder
- delete by id, name, or metadata
- save/load persistence to disk

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ See [`docs/api.md`](docs/api.md) for the full reference.
 - [LangChain](docs/integrations/langchain.md) — `pip install turbovec[langchain]`
 - [LlamaIndex](docs/integrations/llama_index.md) — `pip install turbovec[llama-index]`
 - [Haystack](docs/integrations/haystack.md) — `pip install turbovec[haystack]`
+- [Agno](docs/integrations/agno.md) — `pip install turbovec[agno]`
 
 ## Rust
 
--- a/docs/integrations/agno.md
+++ b/docs/integrations/agno.md
@@ -1,0 +1,65 @@
+# Agno integration
+
+`turbovec.agno.TurboQuantVectorDb` implements Agno's `VectorDb` interface backed by an `IdMapIndex`, supporting insert, upsert, search, and delete with O(1) id-based operations.
+
+## Install
+
+```bash
+pip install turbovec[agno]
+```
+
+## Basic usage
+
+```python
+from agno.agent import Agent
+from agno.knowledge import Knowledge
+from agno.knowledge.embedder.openai import OpenAIEmbedder
+from turbovec.agno import TurboQuantVectorDb
+
+vector_db = TurboQuantVectorDb(
+    dim=1536,
+    bit_width=4,
+    embedder=OpenAIEmbedder(),
+)
+
+knowledge = Knowledge(vector_db=vector_db)
+knowledge.load_text("Turbovec compresses vectors to 4 bits per dimension.")
+
+agent = Agent(knowledge=knowledge)
+agent.print_response("What does turbovec do?")
+```
+
+## Parameters
+
+| Parameter | Type | Default | Description |
+|-----------|------|---------|-------------|
+| `dim` | `int` | `1536` | Embedding dimension |
+| `bit_width` | `int` | `4` | Quantization bits (2 or 4) |
+| `embedder` | `Embedder` | `None` | Agno embedder for search queries |
+| `path` | `str` | `None` | Directory for save/load persistence |
+| `similarity_threshold` | `float` | `None` | Minimum score to include in results |
+
+## Save / load
+
+```python
+vector_db = TurboQuantVectorDb(dim=1536, bit_width=4, path="./my-store")
+vector_db.create()  # loads from path if exists, else creates new
+
+# ... add documents ...
+
+vector_db.save()  # persists to path
+```
+
+## Delete
+
+```python
+vector_db.delete_by_id("doc-123")
+vector_db.delete_by_name("my-document.pdf")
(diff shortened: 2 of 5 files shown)


RyanCodrai:
Hey! Thanks for opening a PR. I will investigate Agno and see if this is something I want to support. I will also create a contributor guideline document as I think it makes more sense to raise issues where we can dicuss implementation plans first before jumping straight to a PR.

RyanCodrai:
Thanks for putting this together — the wiring against Agno's `VectorDb` interface is real work and the test coverage looks solid for the surface that's implemented. A bit of context on where this sits relative to recently-landed work, then the specific gaps:

PR #22 (just merged) standardized three things across the existing integrations that this PR pre-dates:

1. **Kernel-level filtering via `allowlist=` / `mask=`** on `IdMapIndex.search` and `TurboQuantIndex.search`. Filters resolve to an allowlist *before* scoring; the kernel only ever inserts allowed candidates into the per-query heap. No over-fetch, no recall hit on selective filters, no fewer-than-`k` results.
2. **JSON side-cars with `schema_version`** across all integrations. The `allow_dangerous_deserialization` flag is gone everywhere — loading is safe regardless of file provenance.
3. **Lazy index construction** in the core: `IdMapIndex()` / `IdMapIndex(dim=None, bit_width=4)` constructs an empty index that infers `dim` from the first batch. Matches the no-arg constructor pattern of in-tree reference stores in the other three frameworks.

Specific items in this PR that would need to change:

- **`search` ignores `filters`** (`agno.py:147` onwards). The Known Limitations section calls this out, but it's the main thing we should be doing differently from `pip install agno`'s native backends. The new core API supports it directly: build a handle allowlist from the metadata predicate, pass `allowlist=` to `self._index.search`. The same pattern in `turbovec.haystack` / `turbovec.langchain` / `turbovec.llama_index` is the model to mirror.
- **Pickle side-car** (`agno.py:354-371`). All three other integrations use JSON now, and `from_persist_path`-style loaders no longer require any opt-in flag for safety reasons. The 0-day-of-the-PR risk is that someone calls `TurboQuantVectorDb(path=...)` on an attacker-controlled folder; pickle there is genuinely unsafe.
- **`dim=1536` default** (`agno.py:46`). Bakes in an OpenAI-ada assumption that's outdated and silently truncates / mismatches if the user's embedder produces a different shape. Drop it and use `dim=None` (lazy); commit on the first `insert` from `doc.embedding.shape`. Matches the no-arg pattern the other integrations now expose.
- **Per-document `add_with_ids` loop** (`agno.py:99-118`). Each call invalidates the SIMD-blocked cache, which gets re-materialized on the next search. Batch the whole `documents` list into one `add_with_ids` call.
- **No `Document.embedding is None` error** — silently skipped (`agno.py:99`). The other integrations raise `ValueError(f"Document {doc.id!r} has no embedding...")`. Silent skip masks data-pipeline bugs.
- **No drop-in reference comparison.** The structural diff against `langchain_core.vectorstores.InMemoryVectorStore` / `llama_index.core.vector_stores.SimpleVectorStore` / `haystack.document_stores.InMemoryDocumentStore` was load-bearing in PR #22 for catching semantic divergences. For Agno the closest reference is probably `agno.vectordb.lancedb.LanceDb` (single-machine quantized-friendly) — there are surface details there (e.g., what `get_supported_search_types` is expected to return, async-method semantics, hybrid-search hooks) that should be checked rather than guessed.
- **`async_*` methods just call sync** (`agno.py:128`, etc). The other integrations use the embedder's async methods (`aembed_query`) where available. Lower priority but worth aligning.
- **`content_hash_exists` is an O(n) scan** (`agno.py:90-93`). Maintain a `set` of seen hashes — same shape as the `_str_to_u64` map.

I'm going to take a pass at this on `feat/agno-integration` (branched off main) using the same drop-in-audit methodology — pulling LanceDb (or whatever ends up being the closest Agno reference) as the structural template, threading allowlist filtering into search, swapping pickle for JSON, lazy `dim`. It's not closing this PR — the work you did on the protocol method coverage (`name_exists`, `id_exists`, `delete_by_name`, `delete_by_metadata`, `upsert_available`, `get_supported_search_types`, etc.) is genuinely useful and I'll be referencing your implementation for those. I'll comment back here once I have something to compare against.

If you'd rather rework this PR yourself based on the audit above instead of letting it diverge, happy to defer — just let me know.

PR #28, opened 2026-05-17
Add distortion (MSE) benchmark verifying paper bounds

### LINKED ISSUE #27: Add distortion (MSE) benchmark verifying paper bounds
The benchmark suite has recall, speed, and compression benchmarks but no direct measurement of quantization distortion (MSE).

The paper (arXiv:2504.19874, Section 4.4) reports specific MSE values per bit width and proves an upper bound of D_mse <= (3*pi/2) * (1/4^b). A benchmark script that measures actual MSE and compares against these bounds would validate the implementation and provide a reference for users evaluating compression quality.

Proposed: benchmarks/suite/distortion.py that measures self-retrieval score degradation across dimensions (768, 1536, 3072) and bit widths (2, 3, 4), reporting measured MSE vs paper bounds.

raghavenderreddygrudhanti:
Fixes #27

Adds a benchmark measuring quantization distortion (MSE) and comparing against the theoretical bounds from the TurboQuant paper (arXiv:2504.19874, Section 4.4).

## Results (d=1536, 10K random unit vectors)

| Bit Width | Measured MSE | Paper Bound | Ratio | Within Bound |
|-----------|-------------|-------------|-------|--------------|
| 2-bit | 0.235 | 0.295 | 0.80x | YES |
| 3-bit | 0.069 | 0.074 | 0.94x | YES |
| 4-bit | 0.020 | 0.018 | 1.06x | NO (6% over) |

2-bit and 3-bit are within the theoretical upper bound D_mse <= (3*pi/2) * (1/4^b). 4-bit exceeds it by 6%, likely due to finite-dimension correction (the bound assumes d -> infinity).

MSE is stable across dimensions (768, 1536, 3072), confirming the paper's prediction that distortion is dimension-independent.

## Method

Self-retrieval scoring: for unit vector x, MSE = 2*(1 - score) where score is the inner product returned by index.search(x, k=1). No dequantization API needed.

## Files

- benchmarks/suite/distortion.py - benchmark script
- benchmarks/suite/distortion_chart.py - chart generator
- benchmarks/results/distortion.json - raw results
- docs/distortion.svg - chart


RyanCodrai:
Thanks for picking this up — paper-bound validation is exactly the kind of thing that should be in the project, and the motivation in #27 is right.

Two things to flag, though:

## 1. The validation already exists

[`turbovec/tests/distortion.rs`](https://github.com/RyanCodrai/turbovec/blob/main/turbovec/tests/distortion.rs) has five tests covering this ground:

- `codebook_mse_matches_paper_at_high_dim` — analytical, no sampling. Numerically integrates the Lloyd-Max codebook MSE against the `Beta((d-1)/2, (d-1)/2)` PDF and asserts within 5% of Theorem 1.
- `codebook_mse_within_shannon_factor` — asserts the MSE / Shannon-bound ratio is in `(1.0, 3.0)` across bits ∈ {2, 3, 4} × dims ∈ {256, 768, 1536}, directly testing the paper's "within ~2.7× of Shannon" claim.
- `pipeline_deficit_matches_theorem1` — end-to-end empirical: encode → pack → SIMD-score, asserts `1 − mean_self_score` matches Theorem 1's `[(2, 0.1175), (3, 0.03454), (4, 0.009497)]` within 10%.
- `self_query_variance_tightens_with_more_bits` — variance must shrink as bit width grows; catches \"bits silently ignored downstream\" regressions.
- `self_query_recall_at_1` — self-query at 4-bit must rank itself top-1 with ≥99% recall.

These run under \`cargo test -p turbovec\` and gate every change.

## 2. The PR's measurement is off by a factor of 2

\`MSE = 2 * (1 − score)\` assumes \`score = ⟨x_unit, x_hat_unit⟩\` — the cosine identity for unit-norm vectors. But \`index.search()\` returns the **asymmetric IP** score, calibrated so that \`score ≈ 1 − d × MSE_per_coord = 1 − Theorem1(b)\` directly. No factor of 2 is needed — the score deficit *is* the MSE in expectation. That's the comment block at the top of \`pipeline_deficit_matches_theorem1\`.

You can see the doubling in the results:

| bits | PR measurement | Theorem 1 | ratio |
|---|---|---|---|
| 2 | 0.235 | 0.1175 | 2.00 |
| 3 | 0.069 | 0.0345 | 2.00 |
| 4 | 0.0195 | 0.0095 | 2.05 |

The \"4-bit 6% over the paper bound\" headline is the factor-of-2 conversion bumping against the bound, not a real overshoot. True 4-bit MSE is ~0.0095 vs the bound of 0.0184 — well within. The \"finite-dim correction\" explanation isn't quite right either: the PR's measured MSEs are *identical* across d=768/1536/3072 (~0.0195 each), and a finite-dim correction would vary with d.

## Closing this PR

If surfacing distortion as a chart in the README alongside Recall / Compression / Speed has value, that's worth doing — but it should *render* the existing tested values rather than introduce a separate measurement. Different scope, different PR. I'll leave #27 open in case the surfacing piece is something we want to pick up.

Thanks again for the engagement on this — and for the issue write-up, which made the gap easy to assess.

PR #42, opened 2026-05-21
test: add regression tests for lazy-index dim behavior

Kartalops:
## Summary
- Add three regression tests for lazy-index dimensionality behavior.

## Problem
The `TurboQuantIndex(dim=None, ...)` constructor creates a "lazy" index where the dimensionality is committed on first `add()` rather than at construction. There were no existing tests covering:
1. That `dim` is `None` before the first add.
2. That the first add locks in the dimensionality.
3. That a mismatched-dim second add raises.

## Solution
Added `test_lazy_index_dim_is_none_before_add`, `test_lazy_index_dim_commits_to_input_shape_on_first_add`, and `test_lazy_index_rejects_mismatched_dim_on_second_add` to the existing test suite.

## Tests
Existing tests pass; new tests can be verified with: `pytest turbovec-python/tests/test_index.py -v`

## Risk / Compatibility
- Tests only — no production code changes.
- No new dependencies.
- No API change.

### DIFF
--- a/turbovec-python/src/lib.rs
+++ b/turbovec-python/src/lib.rs
@@ -113,6 +113,15 @@ impl TurboQuantIndex {
         self.inner.len()
     }
 
+    fn __repr__(&self) -> String {
+        format!(
+            "turbovec.TurboQuantIndex(dim={:?}, bit_width={}, n_vectors={})",
+            self.inner.dim_opt(),
+            self.inner.bit_width(),
+            self.inner.len()
+        )
+    }
+
     /// Vector dimensionality. Returns ``None`` when the index was
     /// constructed lazily (no ``dim=``) and hasn't seen an add yet;
     /// otherwise an ``int``.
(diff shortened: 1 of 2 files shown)


RyanCodrai:
Thanks for the contribution @okwn — and the third test here actually surfaced a real bug. The dim-mismatch panic was crossing the FFI as `pyo3_runtime.PanicException` (subclass of `BaseException`, not `Exception`), which is why `pytest.raises(Exception)` couldn't catch it. The right fix isn't to widen the exception class in the test — it's to stop panicking on user input.

#43 converts the user-input asserts on the add path to a proper `AddError` type and surfaces it in Python as a catchable `ValueError`. That PR also lands a test of the new error surface, so this contribution's signal is preserved.

The other two tests (`dim is None before add`, `dim commits on first add`) duplicate coverage that already exists in the wrapper suites — each of `test_langchain.py`, `test_llama_index.py`, and `test_haystack.py` exercises the lazy-commit path through wrapper-specific code that ends up calling the same core `add_2d`.

Appreciate you sending this — the binding-error surface was something I'd missed.

PR #48, opened 2026-05-21
test: add empty-index and zero-dim edge case tests for TurboQuantIndex

Kartalops:


RyanCodrai:
Closing this — still no description and no linked issue.

That said, the edge cases you probed here are genuinely interesting and surfaced a real bug. I've filed:

- #50 — `TurboQuantIndex(dim=0)` and `IdMapIndex(dim=0)` are silently accepted because the existing `assert!(dim % 8 == 0)` passes vacuously for zero. Followed by a divide-by-zero panic on the first `add()`. Real bug. Your test's intuition was right; the assertion (`with pytest.raises(ValueError)`) was wrong because today the constructor doesn't raise — it silently succeeds.
- #51 — the eager-empty-then-search Python test gap (with the correct expected shape `(1, 0)`, not `(1, 3)`).

So the substance of this PR isn't being thrown away; it's just not landing as the PR you submitted. Crediting the signal you surfaced, holding the line on the process: see #46 / [CONTRIBUTING.md](https://github.com/RyanCodrai/turbovec/blob/main/CONTRIBUTING.md). Please open an issue first if you want to tackle either #50 or #51 yourself.

PR #58, opened 2026-05-21
Add Python tests for search on eager-empty index

### LINKED ISSUE #51: Missing Python test: search on eager-empty TurboQuantIndex
## Context

The Python test suite covers search-on-empty for the *lazy-uncommitted* case indirectly (via the wrapper suites in test_langchain/test_llama_index/test_haystack, each of which constructs a lazy store and queries before any writes). The Rust suite covers it directly ([turbovec/tests/lazy_init.rs::search_on_lazy_uncommitted_returns_empty](turbovec/tests/lazy_init.rs)).

The **eager-empty case** — `TurboQuantIndex(dim=128, bit_width=4)` then `search()` with no `add()` in between — isn't tested at the Python layer.

## Current behavior

Verified locally:
```python
>>> idx = TurboQuantIndex(dim=128, bit_width=4)
>>> scores, indices = idx.search(np.zeros((1, 128), dtype=np.float32), k=3)
>>> scores.shape, indices.shape
((1, 0), (1, 0))
```

`effective_k = 0` because there are no vectors. Shape is `(nq, 0)`, which is consistent with the Rust side (where `SearchResults.k` is 0 in the same situation). This is sensible behavior; it's just not pinned by a Python test.

## Proposal

Add a small test to `turbovec-python/tests/test_index.py`:

```python
def test_search_on_empty_eager_index_returns_zero_effective_k():
    idx = TurboQuantIndex(dim=128, bit_width=4)
    q = unit_vectors(1, 128)
    scores, indices = idx.search(q, k=3)
    assert scores.shape == (1, 0)
    assert indices.shape == (1, 0)
```

Plus the same for `IdMapIndex`. A handful of lines total.

## Scope

Tiny. Surfaced by [#48](https://github.com/RyanCodrai/turbovec/pull/48), whose proposed test asserted shape `(1, 3)` (which is impossible — there are zero vectors to return). The contributor's *intuition* about an untested edge case was right; the assertion was wrong.

samarthpatel24:
## Related issue
  
  Closes #51

  ## Summary

  - Add `test_search_on_empty_eager_index_returns_zero_effective_k` to `test_index.py` for `TurboQuantIndex`
  - Add the same test to `test_id_map.py` for `IdMapIndex`
  - Both assert shape `(nq, 0)` when searching an eager-constructed index with no vectors added

  ## Motivation

  Surfaced by #48. The lazy-uncommitted search path is already tested via the integration suites, but the
  eager-empty case (`TurboQuantIndex(dim=N, bit_width=B)` → `search()` with no `add()`) had no Python-layer
  coverage.

  ## Test plan

  - [x] `pytest turbovec-python/tests/` passes

### DIFF
--- a/turbovec-python/tests/test_id_map.py
+++ b/turbovec-python/tests/test_id_map.py
@@ -112,3 +112,11 @@ def test_write_and_load_round_trip(tmp_path):
 def test_load_rejects_nonexistent_file():
     with pytest.raises(IOError):
         IdMapIndex.load("/nonexistent/path/does-not-exist.tvim")
+
+
+def test_search_on_empty_eager_index_returns_zero_effective_k():
+    idx = IdMapIndex(dim=128, bit_width=4)
+    q = unit_vectors(1, 128)
+    scores, indices = idx.search(q, k=3)
+    assert scores.shape == (1, 0)
+    assert indices.shape == (1, 0)
--- a/turbovec-python/tests/test_index.py
+++ b/turbovec-python/tests/test_index.py
@@ -170,3 +170,11 @@ def test_add_with_mismatched_dim_raises_value_error():
     idx = TurboQuantIndex(dim=128, bit_width=4)
     with pytest.raises(ValueError, match="dim mismatch"):
         idx.add(unit_vectors(3, 256))
+
+
+def test_search_on_empty_eager_index_returns_zero_effective_k():
+    idx = TurboQuantIndex(dim=128, bit_width=4)
+    q = unit_vectors(1, 128)
+    scores, indices = idx.search(q, k=3)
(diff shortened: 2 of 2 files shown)


RyanCodrai:
Thanks @samarthpatel24 — the change is correct and I appreciate the quick turnaround. Honest framing though: for tiny test-coverage tickets like #51, what's most useful is for someone to weigh in on whether the test gap actually matters (this one is borderline — the behaviour is already correct), not necessarily to land code. I'll be more careful labelling issues as "help wanted" vs. tracking-notes I'll handle myself. Going to close this since I'll fold the tests into the next pass on that file, but please do raise issues if you spot real gaps — those are the most valuable contributions.

PR #82, opened 2026-05-30
Populate Document.id in langchain similarity_search results (issue #81)

### LINKED ISSUE #81: bug: langchain don't return id with documents
hi i notice that langchain document don't return id with documents

RyanCodrai:
## Summary

Closes #81.

`TurboQuantVectorStore._search_vector` constructed the returned `Document` objects without `id=sid`, so `similarity_search`, `similarity_search_with_score`, and `similarity_search_by_vector` all returned `Document.id = None`. Downstream LangChain code (retrievers, chains, evaluation harnesses) couldn't identify search hits by id.

The fix is one kwarg in two places:

- The `Document` passed to a callable filter predicate (so user filters can match on `doc.id`, matching the in-tree `InMemoryVectorStore` convention).
- The `Document` attached to each `(Document, score)` result tuple.

Both call sites already had `sid` (the stable str id) in scope — it just wasn't being passed to the `Document` constructor. The existing `get_by_ids` method already used the right pattern (\`Document(id=sid, page_content=..., metadata=...)\`).

## Tests

Two new regression tests in `test_langchain.py`:

- `test_similarity_search_results_carry_document_id` — asserts all three similarity_search variants return Documents with `.id` set, covering both caller-provided ids and store-generated UUIDs.
- `test_similarity_search_callable_filter_receives_document_id` — asserts the filter callable receives a Document with `.id` set so predicates can filter on it.

Both new tests fail on `main` (verified by inspection — the fix is the `id=sid` addition), pass after the fix. Local run: 39/39 in `test_langchain.py` pass.

## Why CI didn't catch this previously

`test_langchain.py` covered `Document.id` from `get_by_ids` but had no assertions on `Document.id` from `similarity_search` results. The new tests close that gap.

## Test plan

- [x] Local: `pytest tests/test_langchain.py -v` → 39/39 pass
- [ ] CI (the workflow added in #80) runs the same on Linux / macOS / Windows

### DIFF
--- a/turbovec-python/python/turbovec/langchain.py
+++ b/turbovec-python/python/turbovec/langchain.py
@@ -301,7 +301,7 @@ class TurboQuantVectorStore(VectorStore):
             allowed_handles = [
                 self._str_to_u64[sid]
                 for sid, (text, meta) in self._docs.items()
-                if predicate(Document(page_content=text, metadata=dict(meta)))
+                if predicate(Document(id=sid, page_content=text, metadata=dict(meta)))
             ]
             if not allowed_handles:
                 return []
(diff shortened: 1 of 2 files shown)


RyanCodrai:
Superseded by #83, which includes this fix as a cherry-picked commit alongside the broader audit-driven integration work. Closing to avoid duplicate review surface.

Issue #157, opened 2026-07-19
Generator/one-shot iterable inputs consumed twice → misleading crash (llama add, langchain add_documents, haystack write_documents, langchain ids/metadatas/from_texts)


RyanCodrai:
### Summary

Several integration entry points iterate their input container **more than once** without materializing it first, so passing a **generator** (or any one-shot iterable) drains it on the first pass and the subsequent passes see an empty iterator — crashing with a *misleading* error that points at embedding shape / length mismatch / `len()` rather than the real cause (exhausted iterable). Generators are a natural streaming-ingestion input, and the same code is internally inconsistent: the async variants and `add_texts` materialize with `list(...)`, but the sync `add`/`add_documents`/`write_documents` don't.

Found by a systematic bug-hunt pass (wave 13); all confirmed against `main` @ 1e7200c. Severity: crash (borderline silent-data-loss — same anti-pattern). Distinct from #126 (tuple return type) and #127 (wrong dtype). The single fix is `x = list(x)` at each entry point.

### Cases & repros

**1. LlamaIndex `add()` drains `nodes` twice** — `llama_index.py:148` (`[n.node_id for n in nodes]`) then `:159` (`[node.get_embedding() for node in nodes]` sees `[]`):
```python
s.add(mknode(i) for i in range(5))   # ValueError: expected 2D embedding batch, got 1D
```
(`async_add` works — it does `list(nodes)` at `:536`.)

**2. LangChain `add_documents()` drains `documents` 2–3×** — `langchain.py:152-155`:
```python
s.add_documents(Document(page_content=f"t{i}") for i in range(4))
# ValueError: texts, metadatas, and ids must all have the same length
```
(`add_texts` works — it does `list(texts)` at `:108`; the two entry points are inconsistent.)

**3. Haystack `write_documents()` validation drains, then `len()` crashes** — `haystack.py:173` (`any(... for doc in documents)`) then `:191` (`len(documents)`):
```python
store.write_documents(mkdoc(i) for i in range(5))   # TypeError: object of type 'generator' has no len()
```

**4. LangChain generator `ids`/`metadatas` (`langchain.py:115` `len(ids)`) and `from_texts(<numpy array>)` (`:458` `if texts:`)**:
```python
s.add_texts(["a","b","c"], ids=(f"id{i}" for i in range(3)))  # TypeError: ... generator has no len()
[...]

Issue #125, opened 2026-07-18
LangChain load() doesn't validate docs vs str_to_u64 key-set → desynced side-car loads clean, opaque KeyError mid-query


RyanCodrai:
### Summary

The LangChain integration's `load()` validates the handle↔index bijection (via `check_persisted_handles`) but does **not** check that the docstore's `docs` key-set matches `str_to_u64`. A desynced side-car (partial copy, stale backup, manual edit) therefore loads cleanly and fails later with an opaque `KeyError` deep inside a query — the exact late/unclear failure mode `_persist.py` was written to prevent.

Related to #115 (agno skips the handle check entirely); this is a **different gap** — LangChain *does* run the handle check but leaves the docs/`str_to_u64` key-set unvalidated.

Found by a systematic bug-hunt pass (wave 1); confirmed against `main` @ 1e7200c.

### Affected code

- `turbovec-python/python/turbovec/langchain.py:540-547` (`load`), with `_persist.py:21-52` covering only the handle set.
- `KeyError` surfaces at `langchain.py:336` (`self._docs[sid]` inside `similarity_search`).

### Repro

Dump a 2-doc store, delete `state["docs"]["1"]` from `docstore.json` (leave `str_to_u64` intact), reload:

```
load(dir, EMB)                 # succeeds, no error
tq.similarity_search("a", k=2) # KeyError: '1'
```

### Expected vs actual

**Expected:** a clean `ValueError` at load time ("side-car out of sync"), matching the `_persist.py` design intent. **Actual:** load succeeds; a bare `KeyError` is raised mid-query.

### Suggested fix direction

In `load()`, additionally require `state["docs"].keys() == str_to_u64.keys()` and raise the same style of `ValueError` on mismatch.

Issue #124, opened 2026-07-18
LangChain: None inside explicit ids list stored as-is, corrupted to string "null" through dump/load round-trip


RyanCodrai:
### Summary

Passing a `None` inside an explicit `ids` list to the LangChain `TurboQuantVectorStore.add_texts` stores the literal `None` as the document id, and a persistence round-trip (`dump`/`load`) then silently rewrites that id to the **string** `"null"` via JSON serialization. The document's identity changes across persistence, and the return value violates the `list[str]` contract.

Found by a systematic bug-hunt pass (wave 1); confirmed against `main` @ 1e7200c.

### Affected code

- `turbovec-python/python/turbovec/langchain.py:113-116` (`add_texts`; same in `aadd_texts` at :131-136)
- JSON dump/load at `langchain.py:501-515`

### Repro

```python
tq.add_texts(["a", "b"], ids=["x", None])   # returns ['x', None]
tq.dump(d); tq2 = load(d, EMB)
list(tq2._docs.keys())        # ['x', 'null']  <-- 'null' string, not None
tq2.get_by_ids([None])        # []
tq2.get_by_ids(['null'])      # returns the doc
```

The reference `InMemoryVectorStore` (via base `add_texts` + `add_documents`) does a per-entry `doc_id or str(uuid.uuid4())`, so it returns `['x', '<uuid4>']` and never stores `None`.

### Expected vs actual

**Expected:** each `None` entry replaced per-entry with a generated UUID (reference behavior); returned ids are all `str`. **Actual:** `None` kept as the key, then JSON-serialized to `"null"`, silently changing document identity across a save/load cycle.

### Suggested fix direction

In `add_texts`/`aadd_texts`, when `ids` is provided: `ids = [i if i is not None else str(uuid.uuid4()) for i in ids]` (the `add_documents` path already applies this fallback for `doc.id`).

RyanCodrai:
A wave-9 persist-round-trip pass found a second trigger of this same root cause (ids not coerced/validated to `str` at add time), worth folding into the fix here so it covers all non-`str` ids rather than just `None`:

**An `int` id silently becomes a `str` across dump/load** (identity change). `add_texts(["hello"], ids=[5])` works in memory with int id `5`, but `dump` keys the docs payload by the id and `load` rehydrates from JSON object keys, so after a round-trip both `_docs` and `_str_to_u64` hold `"5"` (str). A later `get_by_ids([5])`/`delete([5])` with the original int then misses.
[...]

PR #171, opened 2026-07-19
LangChain: normalize and validate inputs at the add boundary (#124, #126, #139, #157)

### LINKED ISSUE #126: LangChain add_texts(ids=<tuple>) returns the tuple unchanged instead of list[str]
### Summary

The LangChain `add_texts` documents a `list[str]` return type but, when called with a **tuple** of `ids`, returns the tuple unchanged instead of a list. Minor contract violation vs the reference store, which always builds and returns a fresh `list`.

Found by a systematic bug-hunt pass (wave 1); confirmed against `main` @ 1e7200c.

### Affected code

- `turbovec-python/python/turbovec/langchain.py:188` (`result_ids = ids`) → `:230` (`return result_ids`)

### Repro

```python
tq.add_texts(["a"], ids=("t1",))   # returns ('t1',)  -> type tuple, not list
```

### Expected vs actual

**Expected:** `['t1']` (a `list[str]`, per the annotation and reference behavior). **Actual:** `('t1',)` (a `tuple`).

### Suggested fix direction

`result_ids = list(ids)`.

### LINKED ISSUE #139: LangChain: add_texts(metadatas=[None]) raises opaque TypeError from dict(None) instead of naming the bad metadata argument
### Summary

`TurboQuantVectorStore.add_texts(..., metadatas=[None])` raises an opaque `TypeError: 'NoneType' object is not iterable` (from `dict(None)`), whereas LangChain's reference `InMemoryVectorStore` raises a field-named error identifying `metadata` as the bad input. Both correctly reject a `None` metadata entry — the issue is purely the error quality (which argument was wrong is not indicated).

Found by a systematic bug-hunt pass (wave 4); confirmed against `main` @ 1e7200c with the latest `langchain-core` (1.4.9). LangChain-only surface — Haystack/LlamaIndex/agno take metadata off `Document`/`TextNode` objects that coerce `None`→`{}` at construction, so the bad input isn't reachable there.

### Affected code

- `turbovec-python/python/turbovec/langchain.py:229` — `self._docs[id_] = (text, dict(meta))`; `dict(None)` raises.

### Repro

```
REF InMemoryVectorStore metadatas=[None]:
  ValidationError: 1 validation error for Document
  metadata  Input should be a valid dictionary [type=dict_type, input_value=None, input_type=NoneType]

TV LangChain metadatas=[None]:
  TypeError: 'NoneType' object is not iterable
```

### Expected vs actual

**Expected:** a clear error naming the offending argument (`metadata`), as the reference does. **Actual:** a bare `'NoneType' object is not iterable` with no indication of which input was bad.

### Suggested fix direction

In `add_texts`/`aadd_texts`, either coerce `metadatas = [m or {} for m in metadatas]` (if a `None` metadata should default to empty), or raise an explicit `ValueError` naming the `metadatas` argument.

RyanCodrai:
## What

Hardens the `TurboQuantVectorStore` add boundary (`turbovec-python/python/turbovec/langchain.py`) so caller-provided containers are normalized/validated once, up front, before any state is touched. Reference-parity standard throughout is langchain-core's in-tree `InMemoryVectorStore`.

- **None ids → UUIDs** (`add_texts`/`aadd_texts`): a `None` entry in an explicit `ids` list is replaced per-entry with `str(uuid.uuid4())`, matching the reference. Previously `None` was stored as the docstore key and a `dump()`/`load()` round-trip silently rewrote it to the string `"null"` via JSON, changing document identity. Part of #124 (None ids; the non-str-id policy — reject vs coerce for e.g. int ids — is deliberately left unchanged pending a maintainer decision).
- **`add_texts` always returns a fresh `list[str]`**: `ids=("t1",)` previously escaped as the return value unchanged (`result_ids = ids`). Closes #126
- **Metadata validated before any mutation**: a non-dict metadata entry (e.g. `None`) now raises `TypeError: metadatas[i] must be a dict, got NoneType` *before* anything runs. Previously the crash was an opaque `'NoneType' object is not iterable` from `dict(None)` mid-loop — critically *after* `index.add_with_ids`, leaving docs / index / `str_to_u64` desynced in memory (repro on main: index=3, docs=1, map=2 after one failed add), and `dump()` persisted the corruption. The reference rejects bad metadata with a field-named validation error; we mirror that. Closes #139
- **One-shot iterables materialized once at each entry point** (`add_texts`/`aadd_texts` `metadatas`+`ids`, `add_documents`/`aadd_documents` `documents`, `from_texts`/`afrom_texts` `texts`), and `from_texts` uses `len()`-based emptiness instead of bare truthiness. Previously generators were drained on the first pass and later passes crashed with misleading errors (length mismatch, `object of type 'generator' has no len()`, numpy ambiguous-truth-value). Part of #157 (LangChain cases only; the llama_index `add()` and haystack `write_documents()` siblings are separate follow-ups).

## Why one PR

All four are the same root cause — inputs consumed/validated lazily, past the point of no return — and the fixes overlap (the id-normalizing comprehension fixes both #124's None case and #126's tuple case). One commit per issue group on the branch.

## Test evidence

- Each fix has regression tests in `turbovec-python/tests/test_langchain.py` (11 new tests), verified to **fail on the unfixed code** and pass after.
- #139 includes the required store-unchanged regression test: after a rejected add, index length, `_docs`, `_str_to_u64`, `_u64_to_str`, `_next_u64`, `similarity_search_with_score` results, and `get_by_ids` are all byte-identical to before.
- Full suite: `turbovec-python/tests/` — **389 passed** (was 378 on main; +11 new).
- Parked behavior confirmed untouched: `add_texts(["h"], ids=[5])` still returns `[5]` as on main.

## Parked

- #124 non-str-id policy (reject vs coerce int/float ids to `str`, incl. the `2` vs `"2"` dump collision) — one-bit design decision for the maintainer.
- #157 llama_index / haystack cases, and the #125/#133/#115 persist-boundary keyset validation family — separate follow-up PRs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,6 +107,13 @@ appears under each surface it touches.
   raised *after* the vectors had been added to the index, leaving the
   docstore and index desynced in memory — and a subsequent `dump()`
   persisted the corruption. (#139)
+- **LangChain: generator / one-shot-iterable inputs are materialized once
+  at each entry point** (`add_texts` / `aadd_texts` `metadatas` and `ids`,
+  `add_documents` / `aadd_documents`, `from_texts` / `afrom_texts`), and
+  `from_texts` tests emptiness via `len()` so a numpy array of texts works.
+  Previously such inputs were iterated more than once — drained on the
+  first pass — producing misleading length-mismatch / `len()` /
+  ambiguous-truth-value errors. (#157)
 
 ### Docs
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,7 +113,9 @@ appears under each surface it touches.
   `from_texts` tests emptiness via `len()` so a numpy array of texts works.
   Previously such inputs were iterated more than once — drained on the
   first pass — producing misleading length-mismatch / `len()` /
-  ambiguous-truth-value errors. (#157)
+  ambiguous-truth-value errors. `delete` / `adelete` get the same
+  treatment: a multi-element numpy array of ids previously crashed on the
+  `if not ids:` emptiness test. (#157)
 
 ### Docs
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,12 @@ appears under each surface it touches.
   rewritten to the string `"null"` by a `dump`/`load` round-trip. (#124)
 - **LangChain: `add_texts` always returns a fresh `list[str]`** — passing
   `ids` as a tuple previously returned the tuple unchanged. (#126)
+- **LangChain: a non-dict metadata entry (e.g. `None`) is rejected with a
+  `TypeError` naming the bad entry, before any state is touched.**
+  Previously the crash was an opaque `'NoneType' object is not iterable`
+  raised *after* the vectors had been added to the index, leaving the
+  docstore and index desynced in memory — and a subsequent `dump()`
+  persisted the corruption. (#139)
 
 ### Docs
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,12 @@ appears under each surface it touches.
 - Added the `Operating System :: Microsoft :: Windows` classifier to the
   PyPI metadata — Windows x64 wheels have shipped since 0.4.3 but the OS
   classifiers listed only Linux and macOS. (#143)
+- **LangChain: a `None` entry in an explicit `ids` list is replaced with a
+  generated UUID** at add time (matching the reference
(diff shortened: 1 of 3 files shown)


RyanCodrai:
Addressed review: `delete()`/`adelete()` had the same truthiness bug as `from_texts` — `if not ids:` crashed on a multi-element numpy array of ids. Fixed with the same materialize-then-len pattern in 6759a38, with a regression test (np.array, generator, and empty-array ids). Full suite now 390 passed. The optional embed-before-validate cost nit is left as-is per the reviewer's note.

Issue #168, opened 2026-07-19
docs/api.md Rust drift: flat add/add_with_ids panic on lazy index (needs add_2d, undocumented); allowlist shape formula wrong for duplicate ids (min(k,len) vs min(k,unique))


RyanCodrai:
Two `docs/api.md`-vs-Rust drifts a downstream Rust user (e.g. the pg_turbovec embedder, #70) hits when porting from the top-level API reference. Both confirmed by compiling+running an external crate against the current `turbovec` path-dep. api.md's preamble says "The Rust API mirrors it," but these two points don't. rustdoc itself is accurate in both cases — this is api.md drift (distinct from #162 rustdoc gaps). Documentation-only.

Found by a systematic bug-hunt pass (wave 19); confirmed against `main` @ 1e7200c.

### 1. Flat `add`/`add_with_ids` panic on a lazy index; api.md doesn't mention the `_2d` requirement

`docs/api.md:44` (`add(vectors)`) and `:91` (`add_with_ids(vectors, ids)`) present the flat methods and say the first call on a lazy index "locks dim." But on the Rust API the flat methods **panic** on a `new_lazy` index — the lazy first-add requires `add_2d` / `add_with_ids_2d`:

```rust
let mut im = IdMapIndex::new_lazy(4).unwrap();
im.add_with_ids(&vecs, &ids).unwrap();
// PANIC: IdMapIndex dim is not set; use add_with_ids_2d(vectors, dim, ids) on the first add ...

let mut idx = TurboQuantIndex::new_lazy(4).unwrap();
idx.add(&vecs).unwrap();
// PANIC: TurboQuantIndex dim is not set; use add_2d(...)
```
(`turbovec/src/lib.rs:244`, `turbovec/src/id_map.rs:86`.) api.md never mentions `add_2d`/`add_with_ids_2d`, so a Rust user porting the documented flat call panics. **Fix:** note in api.md that the Rust lazy first-add uses `add_2d`/`add_with_ids_2d` (the flat `add`/`add_with_ids` require an already-committed dim).

### 2. api.md allowlist shape formula is wrong for duplicate allowlist ids

`docs/api.md:93` states `effective_k = min(k, len(allowlist))`. Actual: the allowlist is deduplicated into a bool mask, so `effective_k = min(k, unique_allowed)`:

```
allowlist = [5,5,5] (len 3, 1 unique), k=10
api.md expects result width 3; actual width = 1
```
[...]

PR #177, opened 2026-07-23
Core rustdoc accuracy: effective-k stride, dead AVX2 epilogue, missing SearchResults docs

### LINKED ISSUE #120: `IdMapIndex::search` doc promises row stride `k`, actual stride is `min(k, len, n_allowed)`
### Summary

`IdMapIndex::search`'s doc comment promises a row stride of `k` (`qi * k .. (qi + 1) * k`), but the actual stride is `effective_k = min(k, len, n_allowed)`. A Rust caller slicing per the doc panics out-of-bounds or misattributes rows.

Found by a systematic bug-hunt pass (wave 1); confirmed by running a repro against `main` @ 1e7200c.

### Affected code

`turbovec/src/id_map.rs:184` (doc comment on `search`).

### Repro

Index with 5 vectors, 2 queries, k=100 → `search` returns 10 scores/ids (stride 5), not the documented layout (which would require 200). Slicing `ids[qi*k..(qi+1)*k]` per the doc panics; if `k < len` but an allowlist shrinks `n_allowed`, the same slicing silently misattributes rows instead.

The data itself is correct: row `qi` is `qi*effective_k .. (qi+1)*effective_k`.

### Notes

- The Python binding is unaffected — it recomputes `effective_k = scores.len()/nq` (`turbovec-python/src/lib.rs:330`).
- `search_with_allowlist`'s doc already states the effective-count behavior; `search`'s doc doesn't.

### Suggested fix direction

Correct the doc comment to state the effective-k stride (cheapest), or return a struct carrying `k`/`nq` like the inner `SearchResults` does.

### LINKED ISSUE #134: Dead code: avx2_block_epilogue (~190-line unused unsafe fn) has no call sites; diverged from live AVX helpers
### Summary

`avx2_block_epilogue` (an ~190-line x86-only `unsafe fn`) is dead code: it has no call sites anywhere in the crate. The live AVX-512BW/AVX2 kernels use `avx2_batch_flush_to_fa` + `avx2_post_flush_heap_update` instead. Not a runtime bug, but a maintenance hazard — it duplicates epilogue logic that has since diverged from the real helpers, and it would emit a `dead_code` warning on x86 builds.

Found by a systematic bug-hunt pass (wave 2); confirmed against `main` @ 1e7200c.

### Affected code

- `turbovec/src/search.rs:889` — `unsafe fn avx2_block_epilogue` (definition only).

### Repro

```
$ grep -n "avx2_block_epilogue" turbovec/src/search.rs
889:...    // the definition; no other line matches
```

Live epilogue helpers used instead: `avx2_batch_flush_to_fa` (search.rs:581/589/665) and `avx2_post_flush_heap_update` (search.rs:604/676).

### Expected vs actual

**Expected:** helpers are either referenced or removed. **Actual:** a fully unused `unsafe fn`, compiled only on x86_64, that would warn under `dead_code` (the crate has no `deny(warnings)` / `[lints]`, so it's non-fatal today) and whose logic has drifted from the real helpers.

### Suggested fix direction

Delete it, or (if intentionally retained for a future path) wire it back in / annotate `#[allow(dead_code)]` with a comment. No behavioral change either way.

### Note

This was the sole confirmed finding in a wave-2 SIMD/unsafe-kernel pass that otherwise ruled out (with runnable repros on aarch64): mask/allowlist evaluation at word/block boundaries, partial-final-block handling, 2/3/4-bit pack↔LUT↔decode consistency, and the AVX-512 odd-`n_byte_groups`/`n_byte_groups==1` tail branches (unreachable because `dim % 8 == 0` forces an even `n_byte_groups`).

### LINKED ISSUE #162: Rust rustdoc gaps: TurboQuantIndex struct + SearchResults (search return type) undocumented, though api.md points Rust users to rustdoc
### Summary

Two public Rust types lack rustdoc doc-comments. This matters because `docs/api.md:9` explicitly delegates the Rust API to rustdoc ("The Rust API mirrors it — see each type's rustdoc for the exact signatures"), so a docs.rs user sees empty descriptions on the crate's primary type and its search return type. (The Python and integration APIs are fully documented — verified both ways via introspection + doc grep.)

Found by a systematic API-doc-completeness bug-hunt pass (wave 16); confirmed against `main` @ 1e7200c. Documentation-only.

### 1. `TurboQuantIndex` struct has no doc-comment (asymmetric with `IdMapIndex`)

- `turbovec/src/lib.rs:115` — `pub struct TurboQuantIndex {` has no `///` summary above it (the only `///` in its body are on *private* fields, which rustdoc doesn't render).
- Contrast `turbovec/src/id_map.rs:44`, which has `/// ID-addressed wrapper around [\`TurboQuantIndex\`].` above `pub struct IdMapIndex`.
- Result: on docs.rs the crate's primary public type shows an empty description.

### 2. `SearchResults` (return type of public `search`) + fields + accessors undocumented

- `turbovec/src/lib.rs:152` — `pub struct SearchResults {` (no `///`); pub fields `scores`/`indices`/`nq`/`k` (`:153-156`) undocumented; pub methods `scores_for_query(&self, qi)` (`:160`) and `indices_for_query(&self, qi)` (`:164`) undocumented.
- `SearchResults` is returned by the public `TurboQuantIndex::search` (`:404`), so a Rust caller has no rendered explanation of what `nq`/`k` mean or how the accessors slice the flat row-major buffers.

### Suggested fix direction

Add a `///` summary above `pub struct TurboQuantIndex`, and document `SearchResults` (struct + fields + both accessors — noting `scores`/`indices` are row-major `nq × k`, and the effective-k caveat from #120). Purely additive rustdoc.

### Verified complete (no action)

Python `TurboQuantIndex`/`IdMapIndex` — every public method/property/param covered in api.md, nothing documented that no longer exists. Integration extension methods (Haystack `update_by_filter`/`get_metadata_*`/etc., LangChain `dump`/`load`/`from_texts`, LlamaIndex `persist`/`from_persist_*`, agno `save`/`optimize`/`update_metadata`) all documented.

RyanCodrai:
Three Rust-core documentation issues, one dead-code removal. No behavior change — the only code delta is deleting an uncalled function and adding one regression test.

## #120 — `IdMapIndex::search` doc promised row stride `k`

**Verified repro on main:** 5 vectors, 2 queries, `k = 100` → `scores.len() == 10` (stride 5), not the 200 the doc's `qi * k .. (qi + 1) * k` formula implies. The inner index clamps to `effective_k = min(k, len, n_allowed)` (`lib.rs` `search_with_mask`).

**Fix (doc-only, per the issue's cheapest option — return type untouched):** the doc now states the stride is `effective_k = min(k, self.len())`, gives the slicing formula in terms of `effective_k`, and tells callers how to recover it (`scores.len() / nq`). Wording follows `search_with_allowlist`'s existing effective-count style.

**Regression test** (no existing test pinned this): `search_row_stride_is_effective_k_when_k_exceeds_len` in `turbovec/tests/id_map.rs` — asserts `scores.len() == nq * min(k, len)`, the derived stride, and per-row self-match at `qi * effective_k`.

## #134 — dead `avx2_block_epilogue` deleted

`grep -rn avx2_block_epilogue` across the repo: only the definition, zero call sites. The live AVX2 epilogue helpers are `avx2_batch_flush_to_fa` and `avx2_post_flush_heap_update`; the dead copy's logic had drifted from them, so it was a hazard to keep. Deleted (~196 lines incl. doc comment). x86 still compiles: `cargo check -p turbovec --target x86_64-apple-darwin` clean, zero warnings.

## #162 — missing rustdoc on `TurboQuantIndex` / `SearchResults`

Purely additive `///` docs:
- `TurboQuantIndex`: struct summary (positional slots, `swap_remove` instability, pointer to `IdMapIndex`).
- `SearchResults`: struct summary + all four pub fields — `scores`/`indices` documented as row-major `nq × k` where `k` is the *effective* per-query count (`min(k, len, n_allowed)`, the #120 caveat), `scores` descending within each row.
- `scores_for_query` / `indices_for_query`: slicing formula and bounds behavior.

Every factual statement was verified by execution: descending order and row-major self-match via the test suite; the accessor panic claim (`qi >= nq` with `k > 0` panics; `k == 0` never panics) via a throwaway test run before commit.

## Verification

- `cargo test -p turbovec`: **144 passed, 0 failed** (143 base + 1 new)
- `cargo check -p turbovec-python`: clean
- `cargo doc -p turbovec --no-deps`: 7 warnings — identical count to main, no new warnings
- `cargo check -p turbovec` (aarch64) and `--target x86_64-apple-darwin`: both zero warnings
- `docs/api.md` untouched: its effective-k statements are already accurate, and the #168 wording there belongs to PR #170 (checked `gh pr diff 170` for overlap — none)

CHANGELOG: `Unreleased → turbovec — Rust crate`, #120/#162 under Fixed, #134 under Removed, matching existing conventions.

Closes #120
Closes #134
Closes #162


RyanCodrai:
Review follow-up, for the record:

- **Fixed in 27aa212** — the lazy-uninitialized path falsified two new sentences: a `new_lazy` index's `search` returns `nq == 0` / empty results (dim unknown, so the row count is too; verified by execution on both `TurboQuantIndex` and `IdMapIndex`). `SearchResults::nq` now carries a one-clause caveat, and `IdMapIndex::search`'s stride-recovery formula (`scores.len() / nq`) is guarded with `nq > 0`. Doc-warning count unchanged (7, the pre-existing baseline); suite 144/0.
- **CHANGELOG conflicts with #170/#176**: all three PRs insert the same `Unreleased` section header, so whichever lands later needs a trivial mechanical rebase of that hunk. Sequencing left to the maintainer; no action taken here.
- **Accessor panic doc, release-mode counterexample**: at `qi >= 2^62`, `qi * k` can wrap in release mode and return row 0 instead of panicking (debug builds do panic). Deliberately not addressed — absurd input, and a real fix means a `checked_mul` code change out of scope for this docs lane.

PR #192, opened 2026-07-25
tests: make state_sequences second-add test actually discriminate a stale blocked cache (#191)

### LINKED ISSUE #191: Test hardening follow-up: state_sequences second_add_after_search_lets_new_vectors_be_found is false-confidence at its 5+1/dim-128 scale
## Summary

Found during the adversarial review of #190 (the #145 lane), by injection: with the blocked-cache invalidation in `add` disabled (`self.blocked = OnceLock::new()` commented out), `turbovec/tests/state_sequences.rs::second_add_after_search_lets_new_vectors_be_found` **passes** — despite its docstring naming exactly that bug. At 5+1 vectors / dim 128 the stale blocked layout still covers slot 5 within its single block, and slot 5's padding code coincidentally outscores the ~0 dot products of the 5 unrelated unit vectors, so the self-query returns the right answer for the wrong reason.

The strengthened test from #190 (`concurrent_search.rs`, 1024+512 / dim 256 — added slots lie beyond the stale cache's block range, so a stale search can never return them) is currently the sole coverage for this bug class.

## Fix direction

Rescale or restructure the state_sequences variant so its discriminating property holds (e.g. make the second add exceed the first add's block coverage, or assert on a property the stale cache cannot coincidentally satisfy), or fold it into the strong test and delete. Verify by the same injection.

RyanCodrai:
## Mechanism (issue #191)

`state_sequences.rs::second_add_after_search_lets_new_vectors_be_found` was false-confidence at its 5+1/dim-128 scale. The blocked layout packs vectors in blocks of 32 slots, padding the final partial block. With the invalidation in `add` disabled (`self.blocked = OnceLock::new()` at `turbovec/src/lib.rs:356` commented out), the stale single-block cache from the first search still *covers* slot 5 within its padded range, and slot 5's padding code coincidentally outscored the ~0 dot products of the 5 unrelated unit vectors — so the self-query returned the right answer for the wrong reason and the test passed despite its docstring naming exactly this bug. The only real coverage was the strengthened #190 test in `concurrent_search.rs`.

## Fix chosen

Structural block-coverage fix (the issue's preferred direction), mirroring #190's insight at small scale:

- First add fills block 0 **exactly** (32 vectors, `BLOCK = 32`) — the warmed cache has no padded slot slack.
- Second add lands at **slot 32**, the first slot of a block the stale cache does not have. A search against the stale layout can never return slot 32, regardless of how padding codes score — so the discriminating property is structural, not score-dependent (score margins at 4-bit quantization were the coincidence that broke the old version).

Test stays sequential and single-threaded; docstring updated to say what it actually discriminates now (and records the history of the old shape). No src/ changes; the #190 test is untouched.

## Bite verification (both directions)

1. **Injection applied** (invalidation in `add` disabled):
   - Old test: **PASSED** (reproduced the false-confidence defect).
   - #190 control (`concurrent_search.rs::add_after_search_invalidates_blocked_cache`): **FAILED** (`left: 7, right: 1024`) — control confirms the injection works.
   - Fixed test: **FAILED** (`left: 19, right: 32 — new vector not findable after second add+search`) — the test now bites.
2. **Injection reverted**: fixed test passes; full `cargo test -p turbovec` green — **163 tests**, matching the post-#190 baseline.

Injection fully reverted before commit; the diff touches only `turbovec/tests/state_sequences.rs`.

No CHANGELOG entry — tests-only change, per repo precedent (#190).

Closes #191

### DIFF
--- a/turbovec/tests/state_sequences.rs
+++ b/turbovec/tests/state_sequences.rs
@@ -52,27 +52,37 @@ fn unit_vectors(n: usize, dim: usize, seed: u64) -> Vec<f32> {
 fn second_add_after_search_lets_new_vectors_be_found() {
     // `add -> search -> add -> search`: the second add must invalidate
     // the blocked cache populated by the first search, so the new
-    // vectors are visible. Existing tests only assert that returned
-    // indices stay in range (idx < len) after the second add — a bug
-    // that excluded the new batch entirely from the blocked layout
-    // would still satisfy that bound but never surface the new vectors.
+    // vector is visible.
+    //
+    // The blocked layout packs vectors in blocks of 32 slots, padding
+    // the final partial block. To make this test discriminate a stale
+    // cache, the first add fills block 0 *exactly* (32 vectors) and the
+    // second add lands at slot 32 — the first slot of a block the stale
+    // cache does not have. A search against the stale single-block
+    // layout can therefore never return slot 32, no matter how the
+    // padding codes score. (An earlier 5+1 version of this test put the
+    // new vector inside the stale block's padded slot range, where the
+    // padding code could coincidentally win the self-query — it kept
+    // passing with the invalidation in `add` disabled; see issue #191.)
     let dim = 128;
     let mut idx = TurboQuantIndex::new(dim, 4).unwrap();
-    let first = unit_vectors(5, dim, 0x5001);
+    let first = unit_vectors(32, dim, 0x5001);
     idx.add(&first);
-    // First search warms the blocked cache.
+    // First search warms the blocked cache (exactly one full block).
     let _ = idx.search(&first[0..dim], 5);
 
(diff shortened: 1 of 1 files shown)


RyanCodrai:
Review verdict CORRECT (injection bite reproduced both directions; boundary probes at 31/33 confirm the exactly-full block-0 fill is the discriminating property). One deferred follow-up for the record: the hardcoded 32 is forced (BLOCK is private to the crate), and if BLOCK ever stops dividing 32 the test silently downgrades from structural to score-coincidence discrimination. Suggested separate change: a one-line cross-reference comment beside `const BLOCK` in turbovec/src/lib.rs ("state_sequences second-add test assumes 32 divides its first-add count"), or a #[doc(hidden)] pub exposure so tests derive it.

Issue #89, opened 2026-06-07
bug: upsert deletes old data before validating new data — data loss on validation failure


michaelxer:
## Bug: Upsert deletes old data before validating new data — data loss on validation failure

**Files:** `langchain.py:170-201`, `llama_index.py:155-179`, `haystack.py:188-224`, `agno.py:300-352`

### Description

All four integrations follow this upsert pattern:

1. Delete old documents with matching IDs
2. Validate new vectors (dimension check, contiguity)
3. Add new vectors

If step 2 raises (e.g., embedding dimension changed between calls), the old data is already deleted and the store is left empty/inconsistent. The validation should happen *before* the delete.

### Reproducer (LangChain)

```python
import numpy as np
from turbovec.langchain import TurboQuantVectorStore
from langchain_core.embeddings import Embeddings

class VarDimEmbeddings(Embeddings):
    def __init__(self):
        self.call = 0
    def embed_documents(self, texts):
        self.call += 1
        dim = 64 if self.call == 1 else 32
        rng = np.random.default_rng(42)
        v = rng.standard_normal((len(texts), dim)).astype(np.float32)
        v /= np.linalg.norm(v, axis=1, keepdims=True) + 1e-9
        return v.tolist()
    def embed_query(self, t): return [0.0] * 64

emb = VarDimEmbeddings()
store = TurboQuantVectorStore(emb)
store.add_texts(["hello"], ids=["my-id"])     # OK, dim=64
try:
    store.add_texts(["world"], ids=["my-id"])  # deletes "my-id", then dim mismatch!
except ValueError:
    pass
assert "my-id" not in store._docs  # BUG: original data is lost
```

### Expected behaviour

If the new data fails validation, the old data should be preserved. Move dimension/NaN/contiguity checks before the delete step, or defer the delete until after `add_with_ids` succeeds.

### Impact

Any upsert where the embedding model changes dimension, returns NaN, or produces non-contiguous arrays will silently destroy existing data.

---

I spotted this while tracing the upsert path across all four integrations. The pattern is consistent — delete first, validate second — but it means any validation failure after the delete silently destroys existing data.

The fix would be to validate vectors (dimension check, contiguity) *before* the delete step, or to defer the delete until after `add_with_ids` succeeds. Each integration handles this slightly differently so it'd need per-file attention.

[...]

PR #208, opened 2026-07-25
fix: layered thread-safety for the four Python integration stores (#161)

### LINKED ISSUE #161: Not thread-safe under concurrent mutation: add+search on a store raises KeyError (index updated before maps, all 4 integrations); rare core 'Already borrowed'
### Summary

Mutating an index/store from one thread while another thread reads it is not thread-safe, in two distinct ways. Core concurrent *search* is fully correct (verified: 0 mismatches over 64,000 concurrent searches), and `_next_u64` increment is de-facto atomic (0 dup handles over 40,000) — so the issues are specifically **concurrent mutation + concurrent read/mutation**. Both are distinct from #121 (GIL/perf) and #147 (fork).

Found by a systematic bug-hunt pass (wave 15); both confirmed against `main` @ 1e7200c.

### 1. Concurrent add + search on a shared integration store → `KeyError` (torn Python-level state)

All four integration stores update the **Rust index first** and then populate their **Python docstore maps in a separate, unlocked loop**. `IdMapIndex.add_with_ids` is a single atomic GIL-held Rust call that makes the new handle *immediately searchable*, but between that call and the map-population loop, a concurrent `search` on another thread can return a handle that resolves to no entry in the side-maps → `KeyError`.

- LangChain (empirically confirmed): write `_store_texts_and_vectors` — `self._index.add_with_ids(...)` at `langchain.py:216` precedes the map loop at `:225-229`; read `_search_vector` crashes at `:335` (`self._u64_to_str[int(handle)]`) or `:336` (`self._docs[sid]`).
- Same index-before-maps ordering (structurally identical, all vulnerable): `haystack.py` add `:245` / maps `:255` / search lookups after `:456`/`:471`; `llama_index.py` add `:183` / maps `:195-208` / `self._nodes[nid]` at `:519`; `agno.py` add `:357` / maps after / lookups after `:571`.

Repro (thread A loops `add_texts` in batches of 80; 3 threads loop `similarity_search` on the same store; `sys.setswitchinterval(1e-7)`):
```
num_errs: 3   first error kind: search
KeyError: 65
  File ".../turbovec/langchain.py", line 335, in _search_vector   # sid = self._u64_to_str[int(handle)]
final desync entries: 0 | docs: 3201 index: 3201
```
Reliable with large batches (wider window); flaky with 1-doc adds. **Final state is consistent** — this is a transient in-flight crash, not permanent corruption.

**Expected vs actual:** a search concurrent with an add sees either the pre- or post-add state and returns valid docs (as the single-threaded sequence always does — 0 errors). **Actual:** a handle is searchable before it is resolvable → `KeyError`.

### 2. Concurrent `add()` on a shared core `TurboQuantIndex` → rare `RuntimeError('Already borrowed')`

8 threads each looping `idx.add(view[i:i+1])` on one shared index, cold process run with sliced-view (non-owning) numpy args:
```
add: (200, 1600, ["RuntimeError('Already borrowed')", x7])
```
`add(&mut self, ...)` (`turbovec-python/src/lib.rs:59`) takes a pyo3 mutable PyCell borrow; a view-argument conversion appears to yield, letting two threads' mutable borrows overlap and surfacing pyo3's borrow conflict. Only on cold/first-run with view args (never warm, never with fresh contiguous arrays). **Integrity always held** (`len == sum(successful adds)`; failed adds raise, none lost). Low priority (rare, non-corrupting).

### Suggested fix direction

Pick one: (a) document all four stores (and the core mutation methods) as **not thread-safe for concurrent mutation** (matches the reference in-memory stores), or (b) guard the add critical section + the search result-translation with a shared `threading.Lock`/`RLock` so the index update and map population commit atomically w.r.t. readers; for #1, populating the Python maps *before* the index add would close the searchable-before-resolvable window (the delete/upsert path still needs the lock to be fully correct). For #2, a retry on `Already borrowed` is safe since no partial state is committed.

RyanCodrai:
Implements the maintainer-ruled design from the #161 thread-safety research (D7): layered thread-safety — (d) writer-only mutex + (b) reader-safe mutation ordering + (c) tolerant readers — in all four Python integration stores (LangChain, Haystack, LlamaIndex, Agno). Full credit to the empirical research report for the design and the measured evidence; the stress suite here is a port of its harness.

## Design (per the research's measured ruling)

1. **(d) Per-store writer-only mutex** — a `threading.RLock` taken by every mutating method (add/write/insert, delete, upsert, update, clear/drop; sync and async bodies) and by every save path (`dump` / `save_to_disk` / `persist` / `save`, so a save always snapshots a consistent store). **Readers take no lock**, preserving the #186 concurrent-read scaling — the research's plain-mutex foil measured 0.97x (a 3.9x regression), which this design avoids. The lock is the only fix for LlamaIndex's *data-losing* duplicate-handle race (pydantic `PrivateAttr` makes `_next_u64 += 1` tearable — 832 docs silently lost in a 4 s 2-writer run) and Agno's 86.5% delete-vs-delete cleanup-scan crash.
2. **(b) Reader-safe mutation ordering** — adds populate the side-car maps **before** the index insert (a search can only learn a handle from the index, so no reachable handle is ever unmapped), with a per-store failure-unwind that rolls back the pre-inserted map entries (and restores upserted entries) so the #89 "a failed add never destroys existing data" guarantee is preserved; deletes remove from the index **first**, maps second. Validation still precedes any mutation (#89/#139 orderings intact).
3. **(c) Tolerant readers** — result translation uses `.get`-and-skip for handles that vanish mid-flight (the delete-straddle: the one race ordering provably cannot close — 9.4% residual under ordering alone in the research), `list()` snapshots before every reader-side dict iteration, and filtered searches retry a stale allowlist then fall back to a non-raising unfiltered-search + post-filter (bounded retry alone left 2/240k in the research; the fallback closes them). Load-time strictness (#178 keyset/handle checks) is untouched — corruption still fails loud at load; tolerance covers only in-flight windows.

**Documented contract** (each integration doc + class docstring): reads run concurrently and scale; writes serialize on a per-store lock; a read overlapping a write sees pre- or post-write state; under heavy churn a search may transiently return fewer than `k`. Honest-unsafe list documented where user-visible: no cross-call atomicity; batch writes not atomic w.r.t. readers; save serializes with writes; embedders/rerankers run outside the lock; multi-process access unsupported.

## Race matrix — before (research's measured rates on main) → after (this PR)

| store | cell | before | after |
|---|---|---|---|
| langchain | add-vs-search | KeyError 0.32% | **0** |
| langchain | churn-vs-search | KeyError 13.4% | **0** |
| langchain | filtered-search-vs-churn | dict-changed-size 10.0% (+KeyError 4.4%) | **0** |
| haystack | churn-vs-search | KeyError ~2.1% | **0** |
| haystack | filtered-search-vs-churn | dict-changed-size 4.3% | **0** |
| haystack | filter_documents-vs-churn | dict-changed-size 1.9% | **0** |
| llama_index | add-vs-search | KeyError 7.7% | **0** |
| llama_index | churn-vs-search | KeyError 31.8% | **0** |
| llama_index | **add-vs-add** | **DATA LOSS: 832 docs rejected (0.78% of adds), 137 lost `_next_u64` increments** | **0 — `_next_u64` exact** |
| llama_index | filtered-search-vs-churn | dict-changed-size 36.3% | **0** |
| agno | delete-vs-delete | dict-changed-size 86.5% | **0** |
| agno | filtered-search-vs-churn | dict-changed-size 13.5% | **0** |

Verified by checking out main's store sources against the new suite: **29 tests fail on unpatched main** (every scenario class bites on at least one store, matching the research matrix — including the llama add-vs-add data-loss cell), and all pass on this branch. Agno's already-tolerant unfiltered reads stay clean on both, as the research measured.

## Perf gate (research methodology: LangChain store, dim 256, M-series arm64; two interleaved runs each)

| metric | main | this PR | delta |
|---|---|---|---|
| add (ms / 256-doc batch) | 5.22 / 5.18 | 5.23 / 5.30 | +1.2% avg |
| search 2k-doc store (µs) | 47.5 / 47.4 | 47.4 / 47.6 | +0.1% |
| search 50k-doc store (µs) | 277.8 / 277.1 | 278.4 / 278.3 | +0.3% |
| 1T search (ops/s) | 3588 / 3596 | 3571 / 3562 | −0.7% |
| 4T search (ops/s) | 13431 / 13519 | 13377 / 13285 | −1.1% |
| **4T/1T scaling** | **3.74x / 3.76x** | **3.75x / 3.73x** | noise |

All deltas within the ±3% run-to-run noise band; the #186 concurrent-read scaling is intact.

## Tests

- `tests/test_store_thread_safety.py` — port of the research harness: 4 stores × 7 scenarios (add-vs-search, churn-vs-search, filtered-search-vs-churn, add-vs-add, delete-vs-delete, add-vs-delete, get-vs-churn), `slow`-marked at ~1.5 s per cell, asserting zero exceptions in every role + `len(index) == len(each map)` + exact `_next_u64` on add-vs-add (the duplicate-handle catcher).
- Deterministic units (all four stores): maps-before-index-add and index-removal-before-map-pop observed via a spy index; #89 unwind preservation (failed add over an upsert restores the previous entry exactly); tolerant skip of a handle deleted mid-search; non-raising stale-allowlist fallback returning correctly filtered results.
- Full suite: **552 passed** (504 on main baseline + 48 new); `test_gil_release.py` stays green.

## Follow-up suggestions (not in this PR)

- A nightly 4-thread scaling floor probe (≥3x on 4 cores), per the research's test plan — regression-prevention infra, kept separate from the fix.

Closes #161

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -137,6 +137,26 @@ appears under each surface it touches.
 
 #### Changed
 
+- **All four integration stores are now safe for concurrent
+  multi-threaded use; writes serialize on a per-store lock.** The
+  LangChain, Haystack, LlamaIndex, and Agno stores adopt a layered
+  design measured in the #161 research: every mutating method
+  (add/write/insert, delete, upsert, update, clear/drop — sync and
+  async) and every save path (`dump` / `save_to_disk` / `persist` /
+  `save`) serializes on a per-store `threading.RLock`; reads take no
+  lock, so concurrent searches keep overlapping and scaling across
+  threads (#186). Adds now populate the side-car maps *before* the
+  index insert (with a failure-unwind preserving the #89
+  "a failed add never destroys existing data" guarantee) and deletes
+  remove from the index *first*, so a search can never surface a
+  handle that doesn't resolve; result translation skips handles whose
+  entries a concurrent delete removed mid-search, and filtered
+  searches retry a stale allowlist and fall back to a non-raising
+  post-filtered search under sustained churn. The resulting contract:
+  a read overlapping a write sees pre- or post-write state, and under
+  heavy concurrent churn a search may transiently return fewer than
+  `k` results. There is still no cross-call atomicity, and
+  multi-process access remains unsupported. (#161)
 - **Haystack `write_documents` under `FAIL`/`NONE` now partial-writes like
   the reference instead of being atomic.** Previously the whole batch was
   validated up front and a `DuplicateDocumentError` left the store
@@ -155,6 +175,22 @@ appears under each surface it touches.
 
 #### Fixed
 
+- **LlamaIndex `add()` no longer loses data under concurrent calls.**
+  `_next_u64 += 1` on a pydantic `PrivateAttr` is not atomic under
+  the GIL, so two concurrent `add()` calls could issue the same
+  handle — the second batch was rejected with an opaque
+  `ValueError: id already present` and its documents silently never
+  stored (measured: 0.78% of adds under 2-thread ingest). Handle
+  issuance now happens under the store's writer lock. (#161)
+- **Concurrent search / retrieval no longer raises transient
+  `KeyError` / `RuntimeError: dictionary changed size during
+  iteration` while another thread writes.** This closes the measured
+  crash classes in all four stores, including Agno's
+  delete-vs-delete cleanup-scan crash and the misleading
+  `KeyError: allowlist contains id(s) not present in index` from a
+  delete racing a filtered search. Load-time validation is
+  unchanged: a corrupt persisted store still fails loudly at load.
+  (#161)
 - **Agno's `TurboQuantVectorDb` accepts `bit_width=3`.** The constructor
   guard rejected 3 even though the core `IdMapIndex` — and the langchain,
   haystack, and llama_index stores built on it — fully support it; the
--- a/docs/integrations/agno.md
+++ b/docs/integrations/agno.md
@@ -139,6 +139,21 @@ Document metadata must be JSON-serializable — same constraint Agno's `LanceDb`
 
 The lifecycle, write, and read methods have async counterparts: `async_create`, `async_drop`, `async_exists`, `async_name_exists`, `async_get_count`, `async_insert`, `async_upsert`, `async_search`. The remaining methods (the `delete_by_*` family, `update_metadata`, `save`, `id_exists`, `content_hash_exists`, `optimize`) are sync-only. When the embedder exposes `async_get_embedding` / `async_get_embeddings_batch_and_usage`, the async paths use it for genuine async embedding generation.
 
+## Thread safety
+
+The store is safe for concurrent multi-threaded use:
+
+- **Reads run concurrently and scale.** `search`, the existence checks, and `get_count` take no lock; the underlying index releases the GIL during scoring, so independent searches from multiple threads overlap and scale.
+- **Writes serialize.** `insert`, `upsert`, the `delete_by_*` family, `update_metadata`, `drop`, and `save` serialize on a per-store lock. The `async_*` variants delegate to the same locked bodies.
+- **A read overlapping a write sees pre- or post-write state** — never a torn one. Under heavy concurrent churn a search may transiently return fewer than `limit` results (hits deleted mid-search are skipped).
+
+What the contract does *not* cover:
+
+- **No cross-call atomicity.** A caller-side check-then-act sequence (`id_exists` then `delete_by_id`) can interleave with other writers. Batch writes are not atomic with respect to readers: a search overlapping an `upsert` can briefly see both the old and new generation of a `content_hash`.
+- **`save` serializes with writes** (so it always snapshots a consistent store); reads may proceed during a save.
+- **The embedder and reranker are invoked outside the store's lock** and must be thread-safe themselves.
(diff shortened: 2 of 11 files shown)


RyanCodrai:
**Record-only note — free-threaded Python (3.13t+) flags from review, for future reference (no code change in this PR):**

1. **Reader-path atomicity leans on runtime dict guarantees.** Tolerant readers rely on `.get`, single-key `pop`, and `list(dict.items())` being internally consistent — guaranteed by the GIL today and by per-object critical sections in current free-threaded CPython, but the latter is documented behavior, not a language guarantee. If it ever weakens, the reader side needs escalation (e.g. the research's candidate (e) RW lock). Writer side is already sound off-GIL via the explicit lock.
2. **Haystack `update_by_filter` mutates `meta` in place** (`data["meta"].update(meta)` — a single C-call update under the writer lock), whereas agno's `update_metadata` uses copy-then-swap (`data["meta_data"] = merged`). Readers snapshot the outer dict but hand out `dict(data["meta"])` copies; under free-threading, copy-then-swap is the more robust idiom to converge on.
3. **LlamaIndex `query` re-reads `self._index`** across the retry/fallback paths; taking a local reference at entry (as agno's `_retrieve` does) would be more robust against a concurrent `clear()` swapping the index mid-query.

Issue #147, opened 2026-07-19
Fork-unsafe: rayon global pool deadlocks any turbovec op in a forked child (breaks multiprocessing/ProcessPoolExecutor/gunicorn --preload/Celery/DataLoader)


RyanCodrai:
### Summary

Once the parent process has performed any turbovec operation that uses rayon (e.g. `.add()`, which uses `par_iter_mut`), rayon's implicit **global** thread pool is initialized. After `os.fork()`, the child inherits that pool's mutex/registry in a locked-but-worker-less state, and the first rayon use in the child (any `add`/`search`) **deadlocks forever** — with no error. The child hangs even on a brand-new index it builds itself, because it is the global pool that is poisoned, not the inherited index object. There is no `os.register_at_fork` / `pthread_atfork` handler anywhere in the package.

This is the default execution path on Linux for much of the ecosystem: `multiprocessing` (fork was the Linux default through 3.13), `concurrent.futures.ProcessPoolExecutor`, gunicorn/uvicorn `--preload` prefork workers, Celery prefork, and PyTorch `DataLoader(num_workers>0)`.

Found by a systematic bug-hunt pass (wave 7); confirmed against `main` @ 1e7200c (deadlock caught by a 20s watchdog).

### Affected code

- `turbovec/src/encode.rs:69` (`par_iter_mut` in the add/encode path) and the `par_iter` sites in `search.rs`; root cause is rayon's global pool with no fork handler.

### Repro

```
# parent builds an index, forks, child searches (watchdog kills at 20s)
PARENT search ok (4, 5)
TIMEOUT: child did not finish in 20s -> DEADLOCK

# multiprocessing, Linux-default "fork" context, parent-built index queried in worker
[fork] TIMEOUT -> DEADLOCK

# Attribution (isolates the fault to rayon, not BLAS):
PARENT_ADD=0 (clean parent, child builds fresh index + searches)  -> CHILD fresh search ok
PARENT_ADD=1 (parent .add() only, child builds FRESH index)       -> DEADLOCK   # add alone poisons the child
numpy Accelerate GEMM in a forked child (parent primed)           -> ok         # BLAS survives fork
turbovec fresh-index search in forked child (parent primed)       -> DEADLOCK
```

### Expected vs actual

**Expected:** a forked child can use turbovec independently (its own index), or fails loudly with a clear message. **Actual:** silent, indefinite hang the moment the child touches turbovec, whenever the parent already used it.

### Suggested fix direction

[...]

Issue #70, opened 2026-05-25
Public Read/Write I/O API + pub TurboQuantIndex::from_parts


gburd:
## Context

I'm the author of [`pg_turbovec`](https://codeberg.org/gregburd/pg_turbovec), a PostgreSQL extension that uses `turbovec` as the kernel for indexed cosine / inner-product / L2 / L1 ANN search at storage sizes 10–20× smaller than pgvector. It's been carrying a small additive vendor patch on top of `turbovec 0.5.0`/`0.6.0` to handle two needs that database extensions hit but the upstream API doesn't currently support:

1. **In-memory deserialisation.** When the index payload is already in RAM (read out of a `bytea` column or a buffer-manager-pinned page), going through `IdMapIndex::load(path)` requires a tmpfile round-trip — write to `/tmp`, fsync, re-open, read back, parse. On a 1 M × 384-d × 4-bit index that's ≈195 MB of needless I/O. The internal parsers in `src/io.rs` are already generic over `Read`/`Write`; widening the public surface to expose that internal generality is a 100-line additive change.

2. **Constructing an `IdMapIndex` from already-decoded parts.** Database storage backends sometimes have their own page format (e.g. PostgreSQL relfiles) and want to skip the `.tvim` wire format entirely. Promoting `TurboQuantIndex::from_parts` (and the small `packed_codes()` / `scales()` accessors it pairs with) from `pub(crate)` to `pub` lets external embedders construct the index from their own deserialised bytes.

## Proposed API additions

All additive; no behavioural change to existing APIs; no wire-format change.

```rust
// src/io.rs
pub fn write_to<W: Write>(w: &mut W, …) -> io::Result<()>
pub fn load_from<R: Read>(r: &mut R) -> io::Result<…>
pub fn write_id_map_to<W: Write>(w: &mut W, …) -> io::Result<()>
pub fn load_id_map_from<R: Read>(r: &mut R) -> io::Result<…>

// src/id_map.rs
impl IdMapIndex {
    pub fn write_to_writer<W: Write>(&self, w: &mut W) -> io::Result<()>
    pub fn load_from_reader<R: Read>(r: &mut R) -> io::Result<Self>
}

// src/lib.rs — three previously-pub(crate) items become pub
pub fn TurboQuantIndex::from_parts(dim, bit_width, n_vectors, packed_codes, scales) -> Self
pub fn TurboQuantIndex::packed_codes(&self) -> &[u8]
[...]

PR #210, opened 2026-07-25
In-memory serialization API (to_bytes/from_bytes) + correct pickle/copy semantics for all four stores (#148, #149)

### LINKED ISSUE #148: LlamaIndex store: pickle silently returns an EMPTY store (total data loss) instead of round-tripping or erroring
### Summary

`pickle.dumps`/`pickle.loads` of a LlamaIndex `TurboQuantVectorStore` **appears to succeed** but returns a completely **empty** store — every query then returns `[]`. This is total, silent data loss in exactly the scenarios users pickle for: caching, `multiprocessing`, Ray, Celery — ship a populated store, workers deserialize an empty one, no error. It is the only one of the six turbovec objects that silently succeeds-with-wrong-results on pickle (the core `TurboQuantIndex`/`IdMapIndex` and the agno store all raise a clean `TypeError`).

Found by a systematic bug-hunt pass (wave 7); confirmed against `main` @ 1e7200c.

### Affected code

- `turbovec-python/python/turbovec/llama_index.py:81,94` — `BasePydanticVectorStore` is a pydantic model; the Rust index lives in `_index: Any = PrivateAttr()`. Pydantic cannot pickle it, drops `_index`, and resets the other private attrs to blank defaults.

### Repro

```python
li = TurboQuantVectorStore(bit_width=4)
li.add([3 TextNodes with embeddings])       # _nodes {0,1,2}, index len 3
li2 = pickle.loads(pickle.dumps(li))        # "succeeds" (only: WARNING:root:Removing unpickleable private attribute _index)
# li2._nodes == []  index len 0  _node_id_to_u64 {}  _next_u64 0
# li2.query(...) ids -> []   (orig had a,b,c)
```

### Expected vs actual

**Expected:** a faithful round-trip, or a clear "cannot pickle" error (as the core objects give). **Actual:** a pickle that appears to succeed and yields a valid-looking but entirely empty store.

### Suggested fix direction

Implement `__getstate__`/`__setstate__` on the store that serializes the index via the existing `.tvim` save/load (or a Rust `to_bytes`/`from_bytes`) together with the `_nodes` side-car; at minimum override to raise a clear error instead of silently emptying.

### LINKED ISSUE #149: All integration stores: copy.copy shares the Rust index (mutations bleed across copies); deepcopy raises, so no safe copy exists
### Summary

`copy.copy` on any of the four integration stores returns a **shallow copy that shares the same underlying Rust `_index` (and side-car dicts)**, so mutating the copy silently mutates the original (and vice-versa). Users reaching for `copy.copy` to get an independent store instead get an alias. There is no safe fallback: `copy.deepcopy` raises `TypeError: cannot pickle 'builtins.IdMapIndex' object` (deepcopy uses the pickle path), so the only copy that "works" is the unsafe shared one.

Found by a systematic bug-hunt pass (wave 7); confirmed against `main` @ 1e7200c.

### Affected code

- No `__copy__`/`__deepcopy__` defined on any store: `langchain.py:40`, `haystack.py:44`, `llama_index.py:81`, `agno.py:41`.

### Repro

```python
# langchain
c = copy.copy(s)                 # s._index is c._index -> True ; s._docs is c._docs -> True
c.add_texts(["z"], ids=["z"])    # mutate the COPY
# len(s._index) == 4, s._docs keys == ['a','b','c','z']   <- bled into the original

# haystack
hc = copy.copy(h); hc.write_documents([Document "z"])
# h.count_documents() == 4        <- bled into the original

# llama_index
c = copy.copy(li); c.add([node 99])
# li.query(...) ids -> ['0','1','2','3','99']   <- bled into the original

# deepcopy fallback:
copy.deepcopy(s)  -> TypeError: cannot pickle 'builtins.IdMapIndex' object
```

### Expected vs actual

**Expected:** `copy.copy`/`copy.deepcopy` yield an independent store (or raise clearly). **Actual:** `copy.copy` silently aliases the shared Rust index; `deepcopy` raises — so there is no way to get an independent copy.

### Suggested fix direction

Define `__deepcopy__` that clones the underlying Rust index (needs a Rust-side clone/serialize, e.g. via `.tvim` bytes) plus deep-copies the side-car dicts; and either make `__copy__` do the same or raise, so silent aliasing isn't the default.

RyanCodrai:
Implements the maintainer-ruled D5: a core in-memory serialization API and correct copy/pickle semantics for all four integration stores.

Closes #148. Closes #149.
Delivers the in-memory I/O half of #70 (the `from_parts` half landed in #204; leaving #70 open for the reporter to confirm).

## The bugs (reproduced on this branch's base, b524c4c)

**#148 — LlamaIndex pickle silently returns an EMPTY store.** The Rust index lives in a pydantic `PrivateAttr`; `BaseComponent.__getstate__` drops any private attribute that fails `pickle.dumps` with only a log warning:

```
orig query ids:    ['0', '2', '1']
pickled  _nodes:   []
pickled query ids: []          # total, silent data loss
```

The other three stores raised a clean `TypeError: cannot pickle 'builtins.IdMapIndex' object`.

**#149 — copy.copy bleeds; deepcopy raises.** On all four stores, `copy.copy` shared the Rust index (mutating the copy mutated the original: langchain `len(s._index)` 3→4, haystack `count_documents()` 3→4, llama query grew an id, agno `get_count()` 1→2), and `copy.deepcopy` raised `TypeError` — so no safe copy existed.

### 1. Core (Rust): `to_bytes` / `from_bytes` + generic `Read`/`Write` I/O

- `TurboQuantIndex::to_bytes()` / `IdMapIndex::to_bytes()` emit exactly the v4 `.tv`/`.tvim` file bytes; `from_bytes(&[u8])` mirrors `load` with **exactly** the same validation (version handling, structural + value-level checks, v4 rotation-drift verification, `.tvim` duplicate-id check).
- The io module's internals were already generic over `Read`/`Write`; the path-based fns are now thin wrappers over new generic entry points (`io::write_to`, `io::load_from`, `io::write_id_map_to`, `io::load_id_map_from`, plus `write_to_writer<W>` / `load_from_reader<R>` on both types — the exact surface #70 proposed). **No serialization logic is duplicated**: `write`/`write_id_map` call `write_atomic` around the same generic writer; `load`/`load_id_map` open a `BufReader` and call the same generic reader.
- **Rotation-fingerprint cost** (the #205 review observation): `to_bytes` on the index types goes through `rotation_fingerprint()` → `self.rotation.get_or_init`, reusing the cached matrix like `write` does; only the raw `io::write_to`/`io::write_id_map_to` rebuild it (documented). Verified empirically at dim 4096: `add` (builds rotation) 1.80 s, `to_bytes` 0.07 s (O(dim²) hash only); `from_bytes` pays one rebuild for drift verification, same as `load` (1.93 s).
- Properties pinned by tests: `to_bytes()` == the bytes `write(path)` puts on disk (byte-identical, incl. empty and lazy-uncommitted indexes); `from_bytes(to_bytes(x))` reproduces `x` (search/id parity + re-serialization byte-identical); corrupt/drifted bytes fail with the **same error kind and message** as the same bytes passed through a file and `load` (wrong magic, v1 sentinel, unsupported version, truncation, rotation drift, duplicate `.tvim` ids).
- `IdMapIndex` now derives `Debug` (parity with `TurboQuantIndex`).

### 2. Bindings

`to_bytes() -> bytes` / `from_bytes(data)` classmethods on both pyclasses. Serialization runs under the read lock inside `py.detach` (payloads can be big); `from_bytes` snapshots the buffer before detaching (a `bytearray` can be mutated by another thread once the GIL is released) and parses detached. Corrupt bytes raise `ValueError` — there is no path to blame, so not `OSError`. `cargo check -p turbovec-python`: 0 warnings.

### 3. Stores (all four): `__getstate__` / `__setstate__` / `__deepcopy__` / `__copy__`

- State = the index as its `.tvim` bytes + shallow-copied side-car maps + the plain attributes, **snapshotted under the writer lock** so a pickle overlapping a write captures a consistent index/side-car pair (same guarantee the save paths give the on-disk pair).
- Excluded from state and recreated in `__setstate__`: the per-store `RLock` (locks are not picklable), and for Haystack also the `ThreadPoolExecutor` — a restored/copied store always owns a fresh single-thread executor, even when the original wrapped a caller-provided one (documented).
- LlamaIndex restores through `__init__` so the pydantic machinery (`__pydantic_private__` etc.) is fully initialized on the bare unpickled instance; model fields ride along and are re-validated.
- Agno round-trips the not-yet-`create()`d / dropped state as `index_bytes=None`.

### 4. The `__copy__` design call

`__copy__` is **deliberately identical to `__deepcopy__`** — there is no meaningful shallow copy of a store. Evidence from running the canonical references:

| reference | `copy.copy` | `copy.deepcopy` | `pickle` |
|---|---|---|---|
| langchain `InMemoryVectorStore` | shares store dict (bleeds) | independent | round-trips |
| llama `SimpleVectorStore` | shares data dict (bleeds) | independent | round-trips |
| haystack `InMemoryDocumentStore` | shares (bleeds) | **raises** (`_queue.SimpleQueue`) | **raises** |

The references bleed under `copy.copy` only as the accidental default-object behaviour; no framework code calls `copy.copy` on a store and relies on aliasing (grepped the installed packages). For turbovec, alias-sharing a *Rust* index is precisely the #149 bug the maintainer ruled against ("Expected: `copy.copy`/`copy.deepcopy` yield an independent store"), so the safe, least-surprising choice is an independent copy — cheaper to reason about than a documented-but-surprising alias, and it also fixes deepcopy on the Haystack store where even the reference raises. Note our pickle support now *exceeds* the Haystack reference, which cannot pickle at all in 3.0.

## Evidence

**Headline before/after** (same script, unfixed → fixed):

```
# before                                    # after
pickled query ids: []                       pickled query ids: ['0', '2', '1']
langchain deepcopy raised: TypeError        independent copy, both directions
haystack: copy bleeds, count 3→4            original untouched
```

**Biting**: with the store fixes stashed (extension already carrying `to_bytes`), all 16 store tests fail exactly as the issues describe — llama pickle asserts `[] == ['node-0', 'node-1', 'node-2']`, the others fail on `TypeError`/bleed; the 5 core-binding tests pass. With the fixes restored: 21/21 pass.

**Tests added**
- Rust (`turbovec/tests/bytes_io.rs`, 8): `tv_to_bytes_is_byte_identical_to_write_file`, `tvim_to_bytes_is_byte_identical_to_write_file`, `empty_and_lazy_indexes_round_trip_byte_identically`, `tv_from_bytes_round_trip_search_parity`, `tvim_from_bytes_round_trip_search_and_id_parity`, `io_generic_id_map_round_trip_matches_file_load`, `tv_from_bytes_rejects_corrupt_bytes_with_same_errors_as_load`, `tvim_from_bytes_rejects_duplicate_ids_like_load`.
- Python (`turbovec-python/tests/test_pickle_copy.py`, 21): binding round-trips (byte-identity vs file, bytearray, corrupt/wrong-type → `ValueError`/`TypeError`, lazy index), per-store pickle-with-real-data (docs, metadata, score parity, and the handle counter — a post-restore add gets a fresh non-colliding handle), deepcopy independence both directions (the #149 repro), copy.copy no-bleed, a post-unpickle 4-thread × 25-add lock stress, and a `multiprocessing` **spawn**-context smoke (llama + langchain: pickle a populated store into a spawned child and search there; embedders are hashlib-seeded so they're stable across the child's different hash seed). Spawn, not fork — #147 is parked.

**Suites**
- `cargo test -p turbovec`: **216 passed, 0 failed** (208 baseline + 8 new)
- `cargo check -p turbovec-python`: **0 warnings** (`--tests` on the core also 0)
- Full pytest (scratch venv, extension rebuilt): **573 passed, 0 failed** (552 baseline + 21 new)

**Docs**: `docs/api.md` gains `to_bytes`/`from_bytes` rows in both method tables plus an "In-memory serialization" section (covers the Rust generic I/O surface); each integration doc gets a one-line pickle/copy note in its save/load section; CHANGELOG under Unreleased — Rust crate *Added*, Python package *Added* + *Fixed* (llama silent-empty-pickle as the headline).

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -267,9 +267,11 @@ appears under each surface it touches.
   are excluded and recreated on restore), `__deepcopy__`, and
   `__copy__`. The state is snapshotted under the store's writer lock,
   so a pickle overlapping a write captures a consistent index/side-car
-  pair. `pickle.loads(pickle.dumps(store))` round-trips documents,
-  metadata, search results, and the handle counter, including into
-  `multiprocessing` spawn workers. `copy.copy` deliberately equals
+  pair (per-doc payload dicts are copied deeply enough that an
+  in-place metadata update landing mid-pickle cannot tear the
+  snapshot). `pickle.loads(pickle.dumps(store))` round-trips
+  documents, metadata, search results, the similarity mode, and the
+  handle counter, including into `multiprocessing` spawn workers. `copy.copy` deliberately equals
   `copy.deepcopy`: there is no meaningful shallow copy of a store —
   sharing the mutable Rust index means mutations bleed between the
   copies (see *Fixed*). (#148, #149)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,25 @@ appears under each surface it touches.
   version" error (verified against the previous release's loader — no
   silent misparse).
 
+- **In-memory serialization: `to_bytes` / `from_bytes` on both index
+  types, and generic `Read`/`Write` I/O entry points.**
+  `TurboQuantIndex::to_bytes` / `IdMapIndex::to_bytes` serialize an
+  index to its `.tv` / `.tvim` wire format in memory — byte-identical
+  to the file `write(path)` produces — and `from_bytes` mirrors `load`
+  with exactly the same validation (version handling, structural and
+  value-level checks, v4 rotation-drift verification, the `.tvim`
+  duplicate-id check), so bytes and the file they came from load, or
+  fail, identically. `write_to_writer<W: Write>` /
+  `load_from_reader<R: Read>` are the generic-sink forms; the `io`
+  module gains the matching raw entry points `io::write_to`,
+  `io::load_from`, `io::write_id_map_to` and `io::load_id_map_from`.
+  Like the file paths, `to_bytes` on the index types reuses the cached
+  rotation matrix for the v4 fingerprint (no `O(dim³)` rebuild); only
+  the raw `io` writers rebuild it. `IdMapIndex` now derives `Debug`.
+  This delivers the in-memory I/O half of #70 (the `from_parts` half
+  landed in #204) and is the substrate for the Python stores' pickle
+  support. (#148, #149, #70)
+
 #### Changed
 
 - **`MAX_DIM` lowered from 65536 to 16384.** The engine builds a
@@ -233,6 +252,27 @@ appears under each surface it touches.
   importing `importlib.metadata` eagerly would multiply the import
   time by ~25x). Falls back to `"0.0.0.dev0"` when no dist metadata
   is installed. (#153)
+- **`to_bytes()` / `from_bytes(data)` on `TurboQuantIndex` and
+  `IdMapIndex`.** Serialize an index to `bytes` in its `.tv` / `.tvim`
+  wire format — byte-identical to the file `write(path)` produces —
+  and reconstruct it from those bytes with exactly the same validation
+  as `load` (corrupt or drifted payloads raise `ValueError`). Both run
+  with the GIL released; `from_bytes` accepts `bytes` or `bytearray`.
+  This is the in-memory persistence path (caches, databases, pickling)
+  that previously required a filesystem round-trip. (#148, #70)
+- **All four integration stores are picklable and copyable.** The
+  LangChain, Haystack, LlamaIndex, and Agno stores implement
+  `__getstate__` / `__setstate__` (the Rust index rides along as its
+  `.tvim` bytes; the per-store lock — and Haystack's async executor —
+  are excluded and recreated on restore), `__deepcopy__`, and
+  `__copy__`. The state is snapshotted under the store's writer lock,
+  so a pickle overlapping a write captures a consistent index/side-car
+  pair. `pickle.loads(pickle.dumps(store))` round-trips documents,
+  metadata, search results, and the handle counter, including into
+  `multiprocessing` spawn workers. `copy.copy` deliberately equals
+  `copy.deepcopy`: there is no meaningful shallow copy of a store —
+  sharing the mutable Rust index means mutations bleed between the
+  copies (see *Fixed*). (#148, #149)
 
 #### Changed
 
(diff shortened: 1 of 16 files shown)


RyanCodrai:
Rebased onto current main (c6df5d1: #211 similarity modes, #212 MSRV, #213 langchain id validation) and addressed the adversarial-review findings:

1. **Llama similarity-mode pickle fidelity** — `__getstate__` enumerated `model_fields` only, so #211's `_similarity` **PrivateAttr** was dropped and `__setstate__`'s `__init__` call defaulted it to cosine: a `dot_product` store unpickled as cosine with diverging scores. The mode now rides in the state explicitly. Added the missing coverage: mode round-trip (pickle **and** deepcopy) for **both modes on all four stores**, with non-unit embeddings and exact score parity — verified the `[dot_product]` llama case fails on the unfixed code.

2. **Agno snapshot tear** — `__getstate__` shallow-copied `_u64_to_doc` while `update_metadata` mutates the per-doc dicts in place; since pickle serializes after the lock releases, an update landing mid-serialization could tear the snapshot. Per-doc payload dicts (plus `meta_data`/`filters` sub-dicts) are now copied in `__getstate__`. Haystack's `update_by_filter` mutates `data["meta"]` in place identically and got the same fix. Added deterministic snapshot-isolation tests (capture state → in-place update → assert the state still holds pre-update metadata) — both fail on the unfixed code; no wall-clock racing.

Suites on the rebased tree: `cargo test -p turbovec` **216 passed**; `cargo check -p turbovec-python` **0 warnings**; full pytest **628 passed** (597 main baseline measured with this branch's test file excluded + 31 branch tests), including `test_similarity_modes.py`, the store thread-safety stress cells, and the GIL-release tests.

Reviewer's minor note (memoryview unsupported in `from_bytes`) left as documented behavior — `bytes`/`bytearray` accepted.

Issue #165, opened 2026-07-19
LlamaIndex: nested MetadataFilters silently supported (SimpleVectorStore reference raises ValueError); code comment claims 'mirrors SimpleVectorStore'


RyanCodrai:
### Summary

The LlamaIndex integration silently **supports nested `MetadataFilters`** (a `MetadataFilters` inside another's `filter_list`) that the canonical reference `SimpleVectorStore` explicitly **rejects** with `ValueError("Nested MetadataFilters are not supported.")`. turbovec's result for a nested filter is logically correct, but the two stores are therefore **not drop-in interchangeable** for nested filters (identical code raises on `SimpleVectorStore` and succeeds on turbovec, and vice-versa), and the turbovec code comments claim its semantics "mirror SimpleVectorStore" — which this contradicts.

Found by a systematic bug-hunt pass (wave 17); confirmed against `main` @ 1e7200c with llama-index-core 0.14.23. Minor (a parity/claim divergence where turbovec is *more* permissive — not a wrong result set).

### Affected code

- `turbovec-python/python/turbovec/llama_index.py:373-395` — `_filters_match` recurses when `isinstance(f, MetadataFilters)`.
- Reference: `build_metadata_filter_fn` (`llama_index/core/vector_stores/utils.py:163-165`) raises on nested `MetadataFilters`; used by `SimpleVectorStore.query`/`.get_nodes`/`.delete_nodes`.

### Repro

```python
# 4 nodes: n1{age7,NYC} n2{age3,NYC} n3{age20,LA} n4{age7,SF}
nested = AND[ OR[city==NYC, city==LA], NOT[age<5] ]   # depth-2
tv.get_nodes(filters=nested)          # -> ['n1','n3']   (correct nested answer)
sv.query(..., filters=nested)         # -> ValueError: Nested MetadataFilters are not supported.
```

### Expected vs actual

**Expected (per the code's own "mirrors SimpleVectorStore" claim):** same behavior as the reference. **Actual:** turbovec evaluates nested filters (correctly) where the reference raises — so a nested filter validated against one store behaves differently on the other, and the parity claim is false.

### Suggested fix direction

[...]

PR #211, opened 2026-07-25
Per-store similarity modes: cosine (default, normalized) + dot_product (opt-in, raw) (#114)

### LINKED ISSUE #114: Integration docs claim cosine similarity, but the engine returns raw inner products (queries not normalized)
### Summary

The framework integrations document the search score as **cosine similarity in `[-1, 1]`**
and map it to a relevance score via `(sim + 1) / 2`. But the core actually returns the
**raw inner product `⟨v, q⟩`** with both norms intact — the query is never normalized, and
the length-renormalization step restores `‖v‖` on the database side. For embeddings that
aren't unit-normalized, the documented contract is wrong: scores fall outside `[-1, 1]`,
the relevance mapping clamps them, and ranking is by dot-product rather than cosine.

### Where

- `docs/integrations/langchain.md` (~L83–85): *"Scores are raw inner products. Because
  vectors are L2-normalized on insert, inner product equals cosine similarity — higher is
  better, range `[-1, 1]`."*
- `docs/integrations/agno.md` (~L51): *"turbovec stores unit-normalized vectors and the
  kernel's raw score is cosine similarity directly."*
- `langchain.py::_select_relevance_score_fn`, plus the agno/haystack `scale_score`
  mappings — all assume cosine in `[-1, 1]`.

### Evidence

Score scales linearly with the **query** norm:

```python
import numpy as np
from turbovec import TurboQuantIndex
dim = 128
v = np.random.default_rng(1).standard_normal(dim).astype(np.float32); v /= np.linalg.norm(v)
idx = TurboQuantIndex(dim, 4); idx.add(v[None, :].copy())
for name, q in [("unit", v), ("2x", 2*v), ("half", 0.5*v)]:
    s, _ = idx.search(np.ascontiguousarray(q)[None, :], 1)
    print(name, "||q|| =", round(float(np.linalg.norm(q)), 2), "score =", round(float(s[0,0]), 4))
# unit ||q||=1.0 score=1.0013
# 2x   ||q||=2.0 score=2.0026
# half ||q||=0.5 score=0.5006
```

Score also scales with the **database-vector** norm (cosine would be identical for both):

```python
u = np.random.default_rng(2).standard_normal(256).astype(np.float32); u /= np.linalg.norm(u)
db = np.stack([1.0*u, 3.0*u]).astype(np.float32)
idx = TurboQuantIndex(256, 4); idx.add(db.copy())
s, i = idx.search(np.ascontiguousarray(u)[None, :], 2)
print(s[0])   # [3.0043 1.0014]  — not [1.0, 1.0]
```

### Impact

- For unit-normalized embeddings (OpenAI, Cohere, sentence-transformers w/ normalize) the
  score ≈ cosine and everything is fine — which is why this hasn't surfaced.
- For **non-normalized** embeddings: scores exceed `[-1, 1]`, `(sim + 1) / 2` clamps
  everything `> 1` to `1.0` (collapsing all high-relevance scores to the same value), and
  ranking is by dot-product, diverging from `InMemoryVectorStore` — the very store these
  wrappers claim to be drop-in replacements for.

### Suggested fix

Either (a) L2-normalize the query (and document the DB-norm behavior) inside the wrappers
so the cosine contract actually holds, or (b) correct the docs to state the score is a raw
inner product and adjust the relevance mapping / expectations accordingly.

### Environment

- turbovec 0.9.0 (Rust) / 0.8.0 (Python), x86_64 Linux, built from source.

RyanCodrai:
# Per-store similarity modes: `cosine` (default, normalized) + `dot_product` (opt-in, raw) — fixes the broken threshold paths for non-unit embeddings

Implements the maintainer ruling (D6) on #114: each of the four integration stores gains a similarity mode, **fixed at construction and recorded in the persisted side-car**.

- **`cosine` (default).** Documents are L2-normalized at add time and queries at search time, so the engine's raw inner product is true cosine in `[-1, 1]` for embeddings of any magnitude. The existing `(sim + 1) / 2` relevance mappings and every threshold path become correct as-is, and ranking matches each framework's in-tree reference store.
- **`dot_product` (explicit opt-in).** Raw vectors, raw inner-product scores, magnitude-aware ranking — byte-for-byte the previous behavior. Documented dataset-relative-threshold caveat.

**This changes default behavior for non-unit-normalized embedders**: freshly built stores now rank by angle rather than `‖v‖·‖q‖·cosθ`, and score magnitudes move from unbounded to `[-1, 1]`. For unit-normalized embedders (OpenAI, Cohere, sentence-transformers w/ normalize) scores are identical up to quantization noise. Persisted stores are unaffected — see compat below.

## Per-store parameter / compat table

| Store | Mode parameter | Values | Sidecar schema | Legacy sidecar (no mode field) | Mode conflict on load |
|---|---|---|---|---|---|
| LangChain | `similarity` (new keyword, on ctor + `from_texts`/`afrom_texts`; documented turbovec extension per the #165 precedent) | `"cosine"` (default) / `"dot_product"` | `docstore.json` v1 → **v2**, loader accepts (1, 2) | v1 loads as `dot_product` | n/a — `load()` fully reconstructs; the file's mode wins |
| LlamaIndex | `similarity` (new keyword, on ctor + `from_params`; also in `to_dict`/`from_dict` config) | `"cosine"` (default) / `"dot_product"` | `nodes.json` v2 → **v3**, loader accepts (1, 2, 3) | v1/v2 load as `dot_product` | n/a — `from_persist_path` fully reconstructs; the file's mode wins |
| Haystack | `embedding_similarity_function` (existing native parameter, now also governs storage; now runtime-validated) | `"cosine"` (default) / `"dot_product"` | `docstore.json` v2 → **v3** (new `vectors_normalized` field), loader accepts (1, 2, 3) | v1/v2 load with normalization **off** and keep the recorded function for the `scale_score` formula → byte-identical behavior (see note) | n/a — `load_from_disk` fully reconstructs; the file wins |
| Agno | `distance` (existing parameter, mirroring LanceDb's) | `Distance.cosine` (default) / `Distance.max_inner_product` (`Distance.l2` still raises) | `docstore.json` v1 → **v2** (new `distance` field), loader accepts (1, 2) | v1 loads as `max_inner_product`; `self.distance` is updated to reflect it | **`ValueError` at `create()`** — Agno's construct-then-load shape means both sides hold a mode, so a recorded mode conflicting with the constructor's errors instead of silently adopting either |

**Haystack legacy note.** Haystack's sidecar already recorded `embedding_similarity_function` (it only chose the `scale_score` formula pre-D6), so "loads as dot_product" is implemented as a separate `vectors_normalized` flag: a v1/v2 store keeps its recorded function (byte-identical `scale_score`, including the saturating cosine branch it had) while queries and new writes stay raw. Forcing the recorded function to `"dot_product"` instead would have silently switched legacy `scale_score` output from the clamp formula to the sigmoid — not byte-identical.

Unknown mode values raise `ValueError` naming the parameter in all four stores. Zero vectors can't be normalized; they're kept as-is and score `0` against everything, matching the references (LangChain maps the NaN cosine to `0.0`; Haystack substitutes norm 1).

## The three wave-8 threshold breakages — before / after

All reproduced on `main` with non-unit embedders before implementing, now regression tests (`tests/test_similarity_modes.py`).

**1. LangChain `score_threshold` retriever** (all-negative raw IPs −2.4 / −3.6 / −2.94; true cosines −0.2 / −0.6 / −0.98):

```
main            : relevance = [0.0, 0.0, 0.0]  → threshold=0.01 retriever returns []
cosine (default): relevance ≈ [0.40, 0.20, 0.01] → threshold=0.05 returns the two genuine matches, best first
dot_product     : pinned old behavior (relevance all 0.0, retriever returns [])
```

Conversely, raw IPs 380 / 50 / 4 all clamped to relevance 1.0 on main (no threshold could separate them); under the default they're distinct and ordered (≈ 0.975 / 0.75 / 0.55).

**2. Agno `similarity_threshold`** (same geometries):

```
main            : threshold=0.01 discards ALL (all-negative→0.0); threshold=0.99 admits ALL (all-positive-large→1.0)
cosine (default): threshold behaves as a true [0,1] relevance cutoff (0.9 keeps only the genuinely close match)
max_inner_product: pinned old behavior (0.01 → [], 0.99 → everything)
```

**3. Haystack `scale_score=True` on the DEFAULT cosine function** (raw IPs 380 / 50 / 4):

```
main            : [1.0, 1.0, 1.0]              — distinct scores collapse, contract violated
cosine (default): [≈0.975, ≈0.75, ≈0.55]       — distinct, order-preserving
dot_product     : sigmoid(s/100) as before      — unchanged
```

## Reference parity (references actually run)

Cosine-mode ranking on non-unit embeddings (controlled geometry: exact target cosines with gaps ≥ 0.22 — far above 4-bit quantization noise — and magnitudes varying 0.5–30×) matches:

- `langchain_core.vectorstores.InMemoryVectorStore` — exact ranking match,
- `haystack` `InMemoryDocumentStore(embedding_similarity_function="cosine")` — exact ranking match,
- `llama_index` `SimpleVectorStore` — exact ranking match,
- Agno: `LanceDb` was run (lancedb installed locally) but **cannot serve as a cosine oracle** — as of agno 2.8 its wrapper never forwards the distance metric to the LanceDB query builder, so an un-indexed table searches by L2 regardless of `distance=Distance.cosine` (verified: its output exactly matches the analytic L2 ranking). The agno parity test therefore asserts against the analytic cosine ranking, which the other three references agree with on the same geometry.

## Legacy byte-identity

`tests/fixtures/legacy_pre_similarity/` was **generated by running the pre-mode code on `main`** (all four stores persisted with non-unit vectors; `expected.json` records main's exact scores). The regression tests load these fixtures with the new code and assert: mode restored as dot_product/raw, identical ranking, and scores equal to main's recorded values (exact on the writing machine; the committed assertions use `rel=1e-6` only to absorb potential SIMD lane-order rounding across ISAs in CI).

## #208 interplay

Normalization is pure computation and runs where the embedding step runs — outside the writer lock (LangChain `_store_texts_and_vectors` pre-lock, LlamaIndex `add` pre-lock, Agno `insert` pre-lock). Haystack documents arrive pre-embedded and their batch prep already happens under the caller's lock in `_commit_batch`; normalization joins that existing pure-numpy prep. Mutation ordering (maps-before-index-add, index-first deletes, failure unwind) is untouched, and #208's stress suite + the #186 GIL tests stay green (in the full run below).

## Tests

`tests/test_similarity_modes.py` — 35 new tests: the three wave-8 repro groups (cosine fixes them / dot_product pins old behavior), reference parity ×4, legacy fixture byte-identity ×4, mode round-trip through save/load ×4, Agno mode-conflict + legacy adoption, zero-vector handling ×4, unknown-mode validation ×4, kwarg forwarding, `to_dict`/`from_dict` round-trip. **Bite verified:** with the store sources reverted to main (tests + fixtures kept), 29/35 fail; the 6 that pass are exactly the dot_product-pins-old-behavior tests, which is correct by construction.

One existing test updated: `test_agno.py::test_save_writes_json_sidecar` pinned `schema_version == 1` (now 2, and asserts the recorded `distance`).

Full suite: **587 passed** (552 on main baseline + 35 new), including `test_store_thread_safety.py` and `test_gil_release.py`. Rust is untouched (wrapper-layer change only — `git diff` contains no `.rs` files), so `cargo test` was not re-run; the extension was rebuilt from the unmodified sources only to run the Python suite.

## Docs

`docs/integrations/{langchain,agno,haystack,llama_index}.md` each gain a steady-state "Similarity modes" section plus save/load mode-restoration notes. The claims flagged in #114 (langchain.md "inner product equals cosine similarity", agno.md "the kernel's raw score is cosine similarity directly") are now true under the default and have been reworded to describe the mode system. CHANGELOG: Added (the mode), Changed (the default-behavior change for non-unit embedders, stated plainly, + the per-schema compat notes), Fixed (the three threshold breakages).

Closes #114

---


RyanCodrai:
Review correction for the record (verdicts all CORRECT, merging): the PR body's bite-claim says the 6 tests passing on main-reverted sources are 'exactly the dot_product-pins-old-behavior tests' — measured, the actual passers are 2 genuine dot-pins (llama, haystack-sigmoid), 2 zero-vector tests that pass on main for different mechanistic reasons, haystack legacy-writes-stay-raw (main is all-raw anyway), and agno-rejects-l2 (main also rejects); two actual dot-pin tests FAIL on main (TypeError on the new kwarg / main rejecting max_inner_product), which still demonstrates bite. Also recording the reviewer's observation: agno v1 legacy sidecars adopt max_inner_product even over an explicitly passed distance=Distance.cosine (the constructor can't distinguish explicit from default) — matches the D6 ruling's compat row and is documented; flagged for maintainer awareness only.

Issue #152, opened 2026-07-19
LlamaIndex from_persist_dir(namespace=...) path traversal: unsanitized namespace with ../ reads outside persist_dir


RyanCodrai:
### Summary

`TurboQuantVectorStore.from_persist_dir(persist_dir, namespace=...)` interpolates `namespace` straight into the side-car filename with no sanitization, so a `namespace` containing `../` escapes `persist_dir` and reads an arbitrary sibling/parent file. Confirmed by running code (a planted file outside the intended dir was loaded). Severity is bounded — `namespace` is an explicit API argument (default `"default"`), not derived from document data — so this is a defensive-hardening / path-traversal issue, exploitable only if a caller routes untrusted input into `namespace` (or the symmetric `persist_path` write stem).

Found by a systematic defensive-security bug-hunt pass (wave 9); confirmed against `main` @ 1e7200c.

### Affected code

- `turbovec-python/python/turbovec/llama_index.py:673-675` (`from_persist_dir`):
  ```python
  persist_fname = f"{namespace}__vector_store.json"
  persist_path = os.path.join(persist_dir, persist_fname)
  ```
  `os.path.join(persist_dir, "../OUTSIDE__vector_store.json")` resolves outside `persist_dir`. The write side (`persist(persist_path)`) treats the whole arg as a caller-supplied stem under the same caller-trust model.

### Repro

```python
st.persist(str(SB/"OUTSIDE__vector_store.json"))   # write outside the intended dir
inner = SB/"inner"                                  # the intended persist_dir
TurboQuantVectorStore.from_persist_dir(str(inner), namespace="../OUTSIDE")
# -> "Traversal read succeeded, recovered node: ['secret_node'] -> SECRET"
```

### Expected vs actual

**Expected:** the load stays within `persist_dir`, or a `namespace` containing path separators is rejected. **Actual:** it reads an arbitrary path relative to `persist_dir`.

### Suggested fix direction

Reject a `namespace` that contains path separators — e.g. `if os.path.basename(persist_fname) != persist_fname: raise ValueError(...)` — or `os.path.realpath` the resolved path and assert it is within `realpath(persist_dir)`.

### Cleared in the same pass (RuledOut)

[...]

PR #213, opened 2026-07-25
fix: LangChain store loudly rejects non-str ids at the add boundary (#124)

RyanCodrai:
Closes the remainder of #124 (the None-id half shipped in #171).

## What

Per the maintainer ruling on #124: the LangChain store now **loudly rejects non-`str` ids at the add boundary**. In the shared id-normalization path used by `add_texts` / `aadd_texts` (and therefore `add_documents`, `aadd_documents`, `from_texts`, `afrom_texts`), any id that is not `None` and not `str` raises `TypeError` naming the offending id, its type, and its position — before any store mutation. `isinstance(id_, str)` is the only acceptance, so `bool` (an `int` subclass) is rejected too. `None` entries still become generated UUIDs (#171 behavior, pinned by tests).

## Why reject (repro evidence)

The reference `InMemoryVectorStore` accepts-and-corrupts. Verified on unfixed sources:

```
add id=2 (int, 'intdoc') and id="2" (str, 'strdoc')
# in-memory: distinct keys, docs=2, index count=2
dump(): "str_to_u64": {"2": 1, "2": 2}   # json.dump coerces int key 2 -> "2" => duplicate JSON key
load(): last-wins collapse -> 1 entry, one document silently destroyed,
        then ValueError: side-car handle count out of sync with the index -> store unloadable
```

Plus the single-id drift: `add_texts(["hello"], ids=[5])` round-trips as str `"5"`, so a later `get_by_ids([5])`/`delete([5])` misses. Rejecting at the boundary (the declared contract is `list[str]`) makes both states unrepresentable — a deliberate safer-than-reference deviation per the #152 precedent.

## Tests

New in `turbovec-python/tests/test_langchain.py` (all verified failing on unfixed sources — 9 failed / 1 pin passed — and passing with the fix):

- `test_add_texts_rejects_nonstr_id[int|float|bool|bytes|tuple]` — TypeError naming position + type, store unchanged
- `test_add_texts_nonstr_id_mid_batch_leaves_store_unchanged` — mixed batch aborts before any mutation (deep-state assertion: index len, `_docs`, `_str_to_u64`, `_u64_to_str`, `_next_u64`, search results)
- `test_int_str_id_collision_is_unreachable` — the original `2` vs `"2"` corruption scenario can no longer be constructed; surviving doc round-trips intact
- `test_aadd_texts_rejects_nonstr_id`, `test_add_documents_and_from_texts_reject_nonstr_id` — every entry point covered
- `test_add_texts_str_ids_and_none_still_accepted` — happy-path + None→UUID pin

Full suite: **597 passed, 0 failed** (587 baseline post-#211 + 10 new). #208 locking and #211 similarity modes untouched.

## Docs

- CHANGELOG: Python package → Changed (off-contract ids now rejected; was silent corruption through dump/load; safer-than-reference note)
- `docs/integrations/langchain.md`: ids contract documented under "Adding with explicit ids"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -227,6 +227,21 @@ appears under each surface it touches.
 
 #### Changed
 
+- **LangChain: non-`str` ids are now rejected with `TypeError` at the
+  add boundary** (`add_texts` / `aadd_texts` / `add_documents` /
+  `from_texts` and async variants), naming the offending id, its type,
+  and its position, before any embedding-store mutation. Previously an
+  off-contract id (the declared type is `list[str]`) was accepted
+  in-memory and then corrupted by JSON persistence: an `int` id `2`
+  round-tripped as the string `"2"`, and an int id coexisting with its
+  equal-looking str id (`2` + `"2"`) produced a duplicate JSON key on
+  `dump` that `load` collapsed — one document silently destroyed and
+  the side-car left unloadably out of sync with the index. This is a
+  deliberate safer-than-reference deviation: `InMemoryVectorStore`
+  accepts non-str ids and exhibits the same dump/load corruption.
+  `None` entries in an explicit ids list are still replaced with
+  generated UUIDs; `bool` (a subclass of `int`) is rejected like any
+  other non-str type. (#124)
 - **Default scoring of the four integration stores changes for
   non-unit-normalized embeddings.** Under the new `cosine` default the
   stores normalize documents at add time and queries at search time, so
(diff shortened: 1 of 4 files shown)


RyanCodrai:
Review note (verdict CORRECT, merging): one pre-existing non-blocking nitpick for the record — a falsy forced non-str doc.id (Document.model_construct(id=0) / id=False) silently receives a UUID via the pre-existing `doc.id or uuid4` fallback at langchain.py:300/314, rather than the new TypeError. Unreachable without deliberately bypassing pydantic validation; predates this PR.

PR #214, opened 2026-07-25
fix: LlamaIndex dotted-namespace stem collision + legacy fallback; document node_ids=[] semantics (#200, #130)

### LINKED ISSUE #130: LlamaIndex: query(node_ids=[]) returns all nodes instead of none (empty-list truthiness; disagrees with reference, get_nodes, and own docstring)
### Summary

`TurboQuantVectorStore.query(...)` with `VectorStoreQuery(node_ids=[])` (an **empty** list) returns the top-k over **all** nodes, instead of matching nothing. This disagrees with the LlamaIndex reference `SimpleVectorStore`, with turbovec's own `get_nodes`/`delete_nodes` (which use `is not None`), and with the integration's own docstring, which states "Empty list means no node matches."

Found by a systematic bug-hunt pass (wave 2); confirmed against `main` @ 1e7200c with the latest `llama-index-core` (0.14.23).

### Affected code

- `turbovec-python/python/turbovec/llama_index.py:497` — `has_filters = ... or bool(query.node_ids) ...`: `bool([])` is `False`, so an empty `node_ids` skips the filter branch entirely and runs an unfiltered top-k.
- `turbovec-python/python/turbovec/llama_index.py:352` — `_resolve_allowed_handles` uses `if node_ids:` (same truthiness bug), contradicting its own docstring at `:341-342` ("Empty list means no node matches").
- Contrast `get_nodes` (`:292`) and `delete_nodes` (`:239`), which correctly use `if node_ids is not None:`.

### Repro

```python
import numpy as np
from llama_index.core.vector_stores.types import VectorStoreQuery
from llama_index.core.schema import TextNode
from llama_index.core.vector_stores.simple import SimpleVectorStore
from turbovec.llama_index import TurboQuantVectorStore

dim = 16; rng = np.random.default_rng(0)
def mk(i):
    e = rng.standard_normal(dim).astype(np.float32); e /= np.linalg.norm(e)
    n = TextNode(id_=f"n{i}", text=f"t{i}"); n.embedding = e.tolist(); return n
nodes = [mk(i) for i in range(5)]
ref = SimpleVectorStore(); ref.add(nodes)
tq = TurboQuantVectorStore(dim=dim, bit_width=4); tq.add(nodes)
q = rng.standard_normal(dim).astype(np.float32); q /= np.linalg.norm(q)
query = VectorStoreQuery(query_embedding=q.tolist(), similarity_top_k=3, node_ids=[])

print("reference ->", ref.query(query).ids)   # []
print("turbovec  ->", tq.query(query).ids)     # ['n2', 'n0', 'n3']
print("get_nodes ->", [n.node_id for n in tq.get_nodes(node_ids=[])])  # []  (inconsistent w/ query)
```

### Expected vs actual

**Expected:** `query(node_ids=[])` returns an empty result (match nothing), matching the reference `SimpleVectorStore`, turbovec's own `get_nodes(node_ids=[])`, and the `_resolve_allowed_handles` docstring. **Actual:** it returns the top-k over the whole index.

(Note: `doc_ids=[]` was checked and correctly matches the reference — the reference itself treats empty `doc_ids` as no-filter — so this is scoped to `node_ids`.)

### Suggested fix direction

In the `query` path, base the filter decision on `is not None` rather than truthiness (`query.node_ids is not None`), and in `_resolve_allowed_handles` use `if node_ids is not None:` — matching `get_nodes`/`delete_nodes` and the documented semantics.

### LINKED ISSUE #200: LlamaIndex: dotted namespaces silently collide in a shared persist_dir (_split_persist_base strips from the first dot)
## Summary

`_split_persist_base` uses `Path.with_suffix`, which treats everything after the FIRST dot as an extension. A namespace like `v1.2` produces files `v1.tvim`/`v1.nodes.json` — so two dotted namespaces sharing a persist directory (the documented multi-namespace pattern) silently overwrite each other:

```
persist v1.2 and v1.3 in same dir → files: ['v1.nodes.json', 'v1.tvim']
from_persist_dir(namespace="v1.2") → ids ['v1.3']   # v1.3 overwrote v1.2
```

A single dotted-namespace store round-trips fine (write and read mangle identically), which is why this went unnoticed. Found during the adversarial review of #197; pre-existing (not introduced there).

Also noting from the same review, Windows-only and theoretical: the #197 namespace guard doesn't reject drive-relative names like `C:foo` (an escape shape not involving separators or `..`; unreachable on POSIX).

## Fix direction

Replace the `with_suffix`-based stem handling with explicit suffix appending (never strip caller dots), or split only on the known template suffix. Migration note: existing stores persisted under mangled names need a load-time fallback. Optionally extend the namespace guard to reject `:` for the Windows drive-relative shape.

RyanCodrai:
Closes #130
Closes #200

Two ruled LlamaIndex fixes in one component-scoped PR.

## #200 — dotted-namespace stem collision (silent data loss)

`_split_persist_base` and the persist/load paths used `with_suffix`, which strips from the last dot of the stem. Repro (verified on main d3aeaf8):

```
persist v1.2 and v1.3 in the same dir -> files: ['v1.nodes.json', 'v1.tvim']
from_persist_dir(namespace="v1.2")    -> the v1.3 store's data   # v1.3 overwrote v1.2
```

Fix:
- Filenames are built by **appending** `.tvim` / `.nodes.json` to the full namespace-derived stem — `v1.2__vector_store.json` → `v1.2__vector_store.tvim` / `v1.2__vector_store.nodes.json`. Caller dots are never stripped (only the trailing conventional `.json` is). The issue's collision repro now fails to collide; each namespace round-trips its own data.
- **Legacy fallback:** `from_persist_path` falls back to the old mangled names (`v1.tvim` for namespace `v1.2`) when the correct filename is absent — safe because the mangling meant at most one store could survive per mangled prefix. Verified end-to-end against a fixture persisted by **main's actual code** (`git show d3aeaf8:...llama_index.py`): main wrote `['v1.nodes.json', 'v1.tvim']`, branch `from_persist_dir(namespace="v1.2")` loaded the node intact. The next `persist` writes the correct names.
- **Non-dotted namespaces: byte-identical paths before/after** (`default__vector_store.tvim` / `.nodes.json`) — no migration, no fallback involved; pinned by exact-directory-listing test.
- `_validate_namespace` now also rejects `:` anywhere (Windows drive-relative `C:foo` escapes `persist_dir` with no separator); all existing #152/#197 rejections kept.
- The `nodes.json` v3 schema handling (#211) is untouched — the fix is filename-only.

## #130 — `query(node_ids=[])`: KEEP []-means-no-restriction; document reality

Per the maintainer ruling, behavior is kept and the false docstring claim ("Empty list means no node matches") is fixed. Framework evidence (cited in the code): `VectorStoreIndex.as_retriever` passes `node_ids=list(index_struct.nodes_dict.values())` (llama_index/core/indices/vector_store/base.py), and for a `stores_text=True` store `nodes_dict` stays empty — so **every** retriever query arrives with `node_ids=[]` meaning *unrestricted*. Treating `[]` as match-nothing would make every `VectorStoreIndex` retrieval over turbovec return zero results. Verified against llama-index-core 0.14.23.

The docs now tell one coherent story: `query`'s `node_ids`/`doc_ids` follow the retriever calling convention (`[]` = no restriction); `get_nodes` / `delete_nodes` are direct node-selection APIs where `node_ids` *is* the selection (`[]` selects nothing / no-op) — their current behavior, now documented. Updated: `_resolve_allowed_handles`, `get_nodes`, `delete_nodes` docstrings, a comment at `query`'s `has_filters`, and `docs/integrations/llama_index.md`.

## Tests

New in `turbovec-python/tests/test_llama_index.py`:

- `test_dotted_namespaces_coexist_in_shared_persist_dir` — the collision repro; fails on main ("data for v1.3" loaded under namespace v1.2), passes on the branch with unmangled file pairs asserted
- `test_from_persist_dir_loads_legacy_mangled_dotted_store` — legacy fallback (exact mangled byte-paths, verified against main's persist output) + re-persist writes correct names
- `test_non_dotted_namespace_filenames_unchanged` — exact directory listing pin (passes on main and branch: no migration)
- `test_query_empty_node_ids_means_no_restriction` — `node_ids=[]` / `doc_ids=[]` return the unrestricted top-k, `get_nodes([])`/`delete_nodes([])` are empty-selection (pin; passes on main too — behavior kept)
- `test_from_persist_dir_rejects_traversal_namespace` extended with `C:foo` and `a:b` (fail on main, pass on branch); full existing rejection matrix still green
- `test_from_persist_dir_legit_namespace_with_dot_roundtrips` — #197-era collision caveat comment dropped (now false)

Bite verified on unfixed sources: 4 failed (coexist, legacy-fallback setup, `C:foo`, `a:b`), pins passed.

Full suite: **593 passed, 0 failed** (587 baseline post-#211 + 6 new).

## Docs

- CHANGELOG: Python package → Fixed (#200; collision was silent data loss)
- `docs/integrations/llama_index.md`: namespace section rewritten steady-state (truncation caveat dropped — now false; fixed behavior + legacy fallback described), filtered-query section documents the `[]` convention

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -300,6 +300,23 @@ appears under each surface it touches.
 
 #### Fixed
 
+- **LlamaIndex: dotted namespaces no longer silently collide in a shared
+  `persist_dir`.** The persistence stem handling used `with_suffix`,
+  which re-split the stem at its last dot, so namespaces `v1.2` and
+  `v1.3` both persisted to `v1.tvim` / `v1.nodes.json` — the second
+  persist silently overwrote the first (data loss), and
+  `from_persist_dir(namespace="v1.2")` returned the other namespace's
+  data. Extensions are now appended to the full namespace-derived stem
+  (`v1.2__vector_store.tvim` / `v1.2__vector_store.nodes.json`), so
+  dotted namespaces coexist. Non-dotted namespaces keep byte-identical
+  file paths — no migration. A store persisted by an earlier release
+  under a dotted namespace still loads: when the correct filename is
+  absent but the old mangled one exists, `from_persist_path` falls back
+  to it (safe — the mangling meant at most one store could survive per
+  mangled prefix), and the next `persist` writes the correct names.
+  `_validate_namespace` additionally rejects `:` (a Windows
+  drive-relative name like `C:foo` escapes `persist_dir` with no
+  separator), extending the #152/#197 guard. (#200)
 - **Threshold and relevance-score paths are no longer broken for
   non-unit-normalized embeddings** (the three wave-8 findings on #114
   — all symptoms of the mappings assuming cosine input while the
--- a/docs/integrations/llama_index.md
+++ b/docs/integrations/llama_index.md
@@ -144,6 +144,8 @@ Filter semantics match `SimpleVectorStore`'s reference implementation — notabl
 
 Filters are resolved to a handle allowlist **before** scoring. Selective filters return up to `similarity_top_k` matches from the filtered set; you never get fewer just because the filter happened to exclude the top-scoring candidates.
 
+An **empty** `node_ids` (or `doc_ids`) list restricts nothing — it behaves like omitting the argument. This follows the framework's own calling convention: `VectorStoreIndex.as_retriever` passes `node_ids=list(index_struct.nodes_dict.values())`, and for a `stores_text=True` store like this one that list is always empty — it means "unrestricted", and treating it as match-nothing would make every retriever query return zero results. `get_nodes` / `delete_nodes` are different: there `node_ids` *is* the selection, so an explicit empty list selects nothing.
+
 ## Get nodes
 
 ```python
@@ -152,7 +154,7 @@ nodes = vector_store.get_nodes(filters=filters)
 nodes = vector_store.get_nodes(node_ids=["chunk-1", "chunk-2"], filters=filters)  # intersect
 ```
 
-Returns a `List[BaseNode]` reconstructed from the side-car. Missing `node_id`s are silently skipped.
+Returns a `List[BaseNode]` reconstructed from the side-car. Missing `node_id`s are silently skipped. `node_ids` is the explicit selection: an empty list selects nothing and returns `[]` (same for `delete_nodes`, where an empty list is a no-op).
 
 ## Upsert semantics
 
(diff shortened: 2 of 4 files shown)


RyanCodrai:
Review notes for the record (verdicts all CORRECT, merging): (1) direct persist(path) with a non-.json extension changes layout (store.dat → store.dat.tvim, was store.tvim) — verified old-layout stores load via the legacy fallback and migrate on next persist, no data stranded; outside the documented StorageContext convention. (2) A half-present legacy pair (mangled .tvim without its .nodes.json) raises FileNotFoundError naming the mangled sidecar path — already-corrupt state, clean error type, could puzzle a namespace-only user.

PR #57, opened 2026-05-21
SIMD + Rayon encode optimisation (issue #32)

raghavenderreddygrudhanti:
Closes #32.

Profiled encode on ARM/macOS and focused on quantization and memory movement since those dominated wall time.

**Changes:**
- Parallelized normalize using Rayon
- NEON SIMD quantization (boundary counting via u32x4 accumulate)
- Fused quantize + scale + pack — eliminated `codes: Vec<u8>` intermediate
- NEON bit-pack via weighted horizontal add (vaddv)
- x86 falls back to scalar + Rayon
- Verified rotation already multi-threaded through Accelerate BLAS

**Encode throughput (RAYON_NUM_THREADS=1, Apple M2 Pro):**

| Config | main | This PR | Speedup |
|--------|------|---------|---------|
| d=768, 2-bit | 51,925 vec/sec | 108,246 vec/sec | 2.1x |
| d=768, 4-bit | 22,497 vec/sec | 66,310 vec/sec | 2.9x |
| d=1536, 2-bit | 16,602 vec/sec | 25,652 vec/sec | 1.5x |
| d=1536, 4-bit | 9,461 vec/sec | 21,868 vec/sec | 2.3x |

**Recall sweep (all within 0.3pp of main):**

| Dataset | Bit | recall@1 | recall@10 | recall@64 |
|---------|-----|----------|-----------|-----------|
| GloVe-200 | 2-bit | 0.5524 | 0.91 | 0.9957 |
| GloVe-200 | 4-bit | 0.8440 | 0.9997 | 1.0 |
| OpenAI-1536 | 2-bit | 0.877 | 1.0 | 1.0 |
| OpenAI-1536 | 4-bit | 0.956 | 1.0 | 1.0 |
| OpenAI-3072 | 2-bit | 0.919 | 1.0 | 1.0 |
| OpenAI-3072 | 4-bit | 0.966 | 1.0 | 1.0 |

No recall regression. FP order change from fused loop produces identical quantized output at these operating points.

**Search speed (unchanged, as expected — PR only touches encode):**

| Config | ms/query |
|--------|----------|
| d=1536, 2-bit, ARM, ST | 1.252 |
| d=1536, 4-bit, ARM, ST | 2.324 |

**Tests:** All encode tests pass (4/4).

[...]

PR #215, opened 2026-07-25
Add insertion + removal speed benchmarks to benchmarks/suite/ (#65)

### LINKED ISSUE #65: Add insertion + removal speed benchmarks to benchmarks/suite/
## Context

Today \`benchmarks/suite/\` covers **search speed** (e.g. \`speed_d1536_2bit_arm_st.py\`) and **recall** (\`recall_d1536_*.py\`), but not insertion or removal — the only place insert throughput numbers exist is ad-hoc in PR descriptions (e.g. #57's encode SIMD/Rayon work). PR #57 added a quick local Rust bench (\`turbovec/src/bin/encode_bench.rs\`) to produce its 2-3× numbers, then dropped it before merge per the principle that **all performance tests live in benchmarks/suite/**.

This issue captures the gap: we need permanent insertion and removal benches in the canonical location, so future encode-path or remove-path changes have a stable measurement target.

## Proposal

Mirror the structure of the existing speed_* suite. For each (dim, bit_width, arch, threading) cell that we publish search numbers for, add:

- **\`speed_insert_d{DIM}_{BIT}bit_{ARCH}_{THREADING}.py\`** — measures \`TurboQuantIndex.add(vectors)\` throughput in vec/sec, comparing against FAISS PQ insert for the same configuration where applicable. Pin \`RAYON_NUM_THREADS=1\` for ST cells, leave default for MT, matching the convention in the existing search scripts.

- **\`speed_remove_d{DIM}_{BIT}bit_{ARCH}_{THREADING}.py\`** — measures \`IdMapIndex.remove(id)\` throughput (the O(1) swap-and-pop path). Less direct FAISS analogue; could compare against \`IdMapIndex.swap_remove\` on \`TurboQuantIndex\` to show the cost of the id-map layer.

Cells to start with (matching the README's published search-speed cells):

- d=1536, 2-bit and 4-bit, ARM ST/MT, x86 ST/MT
- d=3072, 2-bit and 4-bit, ARM ST/MT, x86 ST/MT

Results land in \`benchmarks/results/\` as JSON, same as the existing scripts.

## Out of scope

- A separate insertion-only diagram in the README — for now, prose mention of insertion throughput is enough. If the numbers are interesting we can revisit.
- Insertion benches at low dim (GloVe-200 etc.) — encode performance is dominated by dim, so 1536/3072 cells are representative. Add if specifically motivated.

## Motivation

After PR #57 lands, encode throughput on ARM at d=1536 is ~22K vec/sec at 4-bit (2.3× faster than before). That's a meaningful win for users building large indices. We should be able to verify the number reproducibly, watch for regressions in future work, and quote it in release notes — none of which is possible without the bench infrastructure.

Surfaced by #57 when its author landed a quick local bench, then we agreed the right home for that measurement is \`benchmarks/suite/\`, not \`turbovec/src/bin/\`.

RyanCodrai:
Closes #65.

Adds permanent insertion and removal speed benchmarks to `benchmarks/suite/`, mirroring the existing `speed_*` cell matrix (d=1536/3072 × 2-bit/4-bit × ARM/x86 × ST/MT): 16 `speed_insert_*` scripts and 16 `speed_remove_*` scripts. Results follow the suite's flat JSON schema (`dim` / `bit_width` / `arch` / `threading` + metric keys) and land in `benchmarks/results/`.

## Methodology

Both operations mutate index state (`add()` is cumulative, `remove()` shrinks the index), so unlike the search scripts the timed op cannot be looped on one index — **every timed run gets a fresh index**, and the reported number is the median of 5 runs, matching the suite's median-of-5 convention. Same deterministic data as the search cells: 100K vectors from `openai-{dim}.npy`, seed 42, unit-normalized. ST cells pin `RAYON_NUM_THREADS=1` (+ `faiss.omp_set_num_threads(1)`); MT cells leave defaults — same as the existing speed scripts.

**`speed_insert_*`** reports four numbers per cell:

- `tq_bulk_insert_vecs_per_sec` — one `add()` of 100K vectors into an empty index. Includes the one-time rotation/codebook init and the TQ+ calibration fit (first add is ≥ 1000 vectors, so calibration is always fitted — never the sub-threshold identity path). This is the full cost a user pays to populate an empty index.
- `tq_warm_insert_vecs_per_sec` — a 10K-vector `add()` onto that 100K index: caches built, calibration frozen by the first add. Isolates the steady-state encode path — this is the regression target for encode-path work like #57.
- `tq_single_add_us` — per-op latency of single-vector `add()` on the warm index (includes the Python-call overhead a caller actually pays).
- `faiss_bulk_insert_vecs_per_sec` — FAISS `IndexPQFastScan` bulk add for the matching configuration (same `m_pq` mapping as the search scripts: `DIM//2` at 2-bit, `DIM` at 4-bit). `train()` is untimed (the one-time analogue of TQ's calibration fit); `reset()` keeps the trained codebooks, so each run times `add()` into an empty trained index.

**`speed_remove_*`** reports three numbers per cell (50K removals in a fixed seeded random order):

- `tq_idmap_remove_us_per_op` / `tq_idmap_removes_per_sec` — `IdMapIndex.remove(id)`: the O(1) swap-and-pop plus the id-map bookkeeping (hash-map remove/insert + `slot_to_id` fix-up).
- `tq_swap_remove_us_per_op` — raw `TurboQuantIndex.swap_remove(slot)` baseline, isolating the id-map layer's overhead. (Both are swap-and-pop: the last vector moves into the freed slot, order is not preserved.)

Timed remove loops include per-call Python overhead — the number is the per-op cost as seen from Python, consistent with how the search scripts measure through the binding.

## ARM results (measured on this machine — Apple M3 Max, same host as the published ARM search cells)

**Insertion**

| Cell | TQ bulk (vec/s) | TQ warm append (vec/s) | TQ single add (µs/op) | FAISS bulk (vec/s) |
|---|---:|---:|---:|---:|
| d=1536 2-bit ST | 21,001 | 80,819 | 65.4 | 45,799 |
| d=1536 2-bit MT | 51,445 | 85,439 | 65.5 | 125,338 |
| d=1536 4-bit ST | 23,679 | 78,900 | 79.9 | 35,065 |
| d=1536 4-bit MT | 33,759 | 81,363 | 99.8 | 83,536 |
| d=3072 2-bit ST | 8,054 | 34,913 | 879.2 | 25,500 |
| d=3072 2-bit MT | 14,289 | 31,355 | 810.6 | 62,122 |
| d=3072 4-bit ST | 5,359 | 19,866 | 1,395.3 | 10,761 |
| d=3072 4-bit MT | 11,808 | 28,962 | 839.8 | 35,914 |

**Removal**

| Cell | IdMap remove (µs/op) | IdMap remove (ops/s) | swap_remove (µs/op) |
|---|---:|---:|---:|
| d=1536 2-bit ST | 0.384 | 2,601,000 | 0.281 |
| d=1536 2-bit MT | 0.727 | 1,375,376 | 0.597 |
| d=1536 4-bit ST | 0.377 | 2,651,137 | 0.267 |
| d=1536 4-bit MT | 0.825 | 1,212,519 | 0.675 |
| d=3072 2-bit ST | 0.234 | 4,278,868 | 0.177 |
| d=3072 2-bit MT | 0.255 | 3,918,534 | 0.210 |
| d=3072 4-bit ST | 0.320 | 3,126,050 | 0.255 |
| d=3072 4-bit MT | 0.339 | 2,953,236 | 0.272 |

Cross-checks on the recorded numbers:

- Bulk ST at d=1536 (21.0K / 23.7K vec/s) is the same order as #57's ad-hoc encode bench (25.7K / 21.9K on an M2 Pro) — that bench measured `encode()` inside Rust; the suite numbers additionally carry the calibration fit + one-time init inside the timed first `add()`, which is why warm append (the pure steady-state encode path) is 3-4× higher.
- Warm append scales ~4× down from d=1536 to d=3072, matching the d² rotation cost per vector.
- Removal is O(1)-flat: sub-µs per op everywhere, with the id-map layer adding a consistent ~0.05-0.15µs of bookkeeping over raw `swap_remove`. The MT remove cells read slightly slower than ST — removal itself is serial; the plausible cause is Rayon workers spin-waiting right after the untimed MT index build. Recorded as measured.
- FAISS bulk add is faster than TQ bulk in most cells — expected: `IndexPQFastScan.add()` encodes against already-trained codebooks with no per-vector d×d rotation. Recorded as measured; no prose claims drawn from it here.

## x86 pending

The 16 `speed_insert_*_x86_*.py` / `speed_remove_*_x86_*.py` scripts are committed runnable but not run — ARM and x86 results have always landed from separate runs. One pass of the new x86 scripts on the GCP bench instance completes the result set.

## Notes

- No README prose changes, no diagram changes (`create_diagrams.py` references speed files explicitly, so nothing breaks; the issue marks an insert diagram out of scope), no CI changes.
- CHANGELOG: entry added under `[Unreleased] → Benchmarks`.
- Full `cargo test` and `pytest` suites run clean (scripts are additive; nothing under test changed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -471,6 +471,19 @@ appears under each surface it touches.
 
 ### Benchmarks
 
+- Insertion and removal speed benchmarks join the suite. (#65) For every
+  published search-speed cell (d=1536/3072 × 2-bit/4-bit × ARM/x86 ×
+  ST/MT), `benchmarks/suite/` gains `speed_insert_*` — bulk `add()` into
+  an empty index (rotation/codebook init + TQ+ calibration fit included),
+  a warm append with calibration frozen (the steady-state encode path),
+  single-vector `add()` latency, and a FAISS `IndexPQFastScan` bulk-add
+  baseline — and `speed_remove_*` — `IdMapIndex.remove(id)` per-op
+  latency and throughput against a raw `TurboQuantIndex.swap_remove`
+  baseline, isolating the id-map layer's bookkeeping. Fresh index per
+  timed run (add is cumulative, remove shrinks the index), fixed seeds,
+  median of 5, results in `benchmarks/results/` like the rest of the
+  suite. ARM (Apple M3 Max) results are recorded; x86 cells await a run
+  on the reference bench instance.
 - `benchmarks/download_data.py` downloads to a `.tmp` sibling and renames
   into place on success, so an interrupted download no longer leaves a
   partial file at the final path that the existence guard then treats as
--- a/benchmarks/results/speed_insert_d1536_2bit_arm_mt.json
+++ b/benchmarks/results/speed_insert_d1536_2bit_arm_mt.json
@@ -1,0 +1,10 @@
+{
+  "dim": 1536,
+  "bit_width": 2,
+  "arch": "arm",
+  "threading": "mt",
+  "tq_bulk_insert_vecs_per_sec": 51445,
+  "tq_warm_insert_vecs_per_sec": 85439,
+  "tq_single_add_us": 65.52,
+  "faiss_bulk_insert_vecs_per_sec": 125338
+}
--- a/benchmarks/results/speed_insert_d1536_2bit_arm_st.json
+++ b/benchmarks/results/speed_insert_d1536_2bit_arm_st.json
@@ -1,0 +1,10 @@
+{
+  "dim": 1536,
+  "bit_width": 2,
+  "arch": "arm",
+  "threading": "st",
+  "tq_bulk_insert_vecs_per_sec": 21001,
+  "tq_warm_insert_vecs_per_sec": 80819,
+  "tq_single_add_us": 65.43,
+  "faiss_bulk_insert_vecs_per_sec": 45799
+}
--- a/benchmarks/results/speed_insert_d1536_4bit_arm_mt.json
+++ b/benchmarks/results/speed_insert_d1536_4bit_arm_mt.json
@@ -1,0 +1,10 @@
+{
+  "dim": 1536,
+  "bit_width": 4,
+  "arch": "arm",
+  "threading": "mt",
+  "tq_bulk_insert_vecs_per_sec": 33759,
+  "tq_warm_insert_vecs_per_sec": 81363,
+  "tq_single_add_us": 99.76,
+  "faiss_bulk_insert_vecs_per_sec": 83536
+}
--- a/benchmarks/results/speed_insert_d1536_4bit_arm_st.json
+++ b/benchmarks/results/speed_insert_d1536_4bit_arm_st.json
@@ -1,0 +1,10 @@
+{
+  "dim": 1536,
+  "bit_width": 4,
+  "arch": "arm",
+  "threading": "st",
+  "tq_bulk_insert_vecs_per_sec": 23679,
+  "tq_warm_insert_vecs_per_sec": 78900,
+  "tq_single_add_us": 79.87,
+  "faiss_bulk_insert_vecs_per_sec": 35065
+}
--- a/benchmarks/results/speed_insert_d3072_2bit_arm_mt.json
+++ b/benchmarks/results/speed_insert_d3072_2bit_arm_mt.json
@@ -1,0 +1,10 @@
+{
+  "dim": 3072,
+  "bit_width": 2,
+  "arch": "arm",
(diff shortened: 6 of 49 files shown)


vishalsachanromi10 (dismissed):

RyanCodrai:
Superseded by a recreated PR (same branch, refreshed benchmark results incoming).

PR #217, opened 2026-07-26
Benchmarks: insertion + removal speed suites (refs #65)

### LINKED ISSUE #65: Add insertion + removal speed benchmarks to benchmarks/suite/

RyanCodrai:
Refs #65.

- [x] ARM cells: scripts + results measured on the reference ARM host (Apple M3 Max) + README figures
- [x] x86 cells: scripts + results measured on the maintainer's GCP bench instance (`c3-standard-8`, Intel Sapphire Rapids) + README figures

Adds permanent insertion and removal speed benchmarks to `benchmarks/suite/`, mirroring the existing `speed_*` cell matrix (d=1536/3072 × 2-bit/4-bit × ARM/x86 × ST/MT): 16 `speed_insert_*` scripts and 16 `speed_remove_*` scripts. Results follow the suite's flat JSON schema (`dim` / `bit_width` / `arch` / `threading` + metric keys) and land in `benchmarks/results/`. Per the maintainer's ruling mid-review, this PR also delivers the README figures (see below).

## Methodology

Both operations mutate index state (`add()` is cumulative, `remove()` shrinks the index), so unlike the search scripts the timed op cannot be looped on one index — **every timed run gets a fresh index**, and the reported number is the median of 5 runs, matching the suite's median-of-5 convention. Same deterministic data as the search cells: 100K vectors from `openai-{dim}.npy`, seed 42, unit-normalized. ST cells pin `RAYON_NUM_THREADS=1` (+ `faiss.omp_set_num_threads(1)`); MT cells leave defaults — same as the existing speed scripts.

**`speed_insert_*`** reports four numbers per cell:

- `tq_bulk_insert_vecs_per_sec` — one `add()` of 100K vectors into an empty index. Includes the one-time rotation/codebook init and the TQ+ calibration fit (first add is ≥ 1000 vectors, so calibration is always fitted — never the sub-threshold identity path). The full cost a user pays to populate an empty index, and the closest analogue of #57's ad-hoc encode bench.
- `tq_warm_insert_vecs_per_sec` — a 10K-vector `add()` onto that 100K index: caches built, calibration frozen by the first add. Isolates the steady-state encode path.
- `tq_single_add_us` — per-op latency of single-vector `add()` on the warm index (includes the Python-call overhead a caller actually pays).
- `faiss_bulk_insert_vecs_per_sec` — FAISS `IndexPQFastScan` bulk add for the matching configuration (same `m_pq` mapping as the search scripts: `DIM//2` at 2-bit, `DIM` at 4-bit). `train()` is untimed (the one-time analogue of TQ's calibration fit); `reset()` keeps the trained codebooks, so each run times `add()` into an empty trained index.

**`speed_remove_*`** reports three numbers per cell (50K removals in a fixed seeded random order):

- `tq_idmap_remove_us_per_op` / `tq_idmap_removes_per_sec` — `IdMapIndex.remove(id)`: the O(1) swap-and-pop plus the id-map bookkeeping (hash-map remove/insert + `slot_to_id` fix-up).
- `tq_swap_remove_us_per_op` — raw `TurboQuantIndex.swap_remove(slot)` baseline, isolating the id-map layer's overhead. (Both are swap-and-pop: the last vector moves into the freed slot, order is not preserved.)

Timed remove loops include per-call Python overhead — the number is the per-op cost as seen from Python, consistent with how the search scripts measure through the binding.

## ARM results (Apple M3 Max — same host as the published ARM search cells)

**Insertion**

| Cell | TQ bulk (vec/s) | TQ warm append (vec/s) | TQ single add (µs/op) | FAISS bulk (vec/s) |
|---|---:|---:|---:|---:|
| d=1536 2-bit ST | 37,427 | 123,499 | 40.5 | 76,400 |
| d=1536 2-bit MT | 139,160 | 143,381 | 40.8 | 464,397 |
| d=1536 4-bit ST | 34,723 | 117,518 | 41.9 | 49,707 |
| d=1536 4-bit MT | 123,714 | 155,943 | 43.0 | 266,005 |
| d=3072 2-bit ST | 12,973 | 51,891 | 356.3 | 36,470 |
| d=3072 2-bit MT | 51,249 | 83,867 | 335.4 | 272,894 |
| d=3072 4-bit ST | 12,578 | 50,904 | 341.9 | 24,250 |
| d=3072 4-bit MT | 49,886 | 95,951 | 342.3 | 164,376 |

**Removal**

| Cell | IdMap remove (µs/op) | IdMap remove (ops/s) | swap_remove (µs/op) |
|---|---:|---:|---:|
| d=1536 2-bit ST | 0.198 | 5,045,027 | 0.174 |
| d=1536 2-bit MT | 0.233 | 4,300,644 | 0.192 |
| d=1536 4-bit ST | 0.222 | 4,499,657 | 0.185 |
| d=1536 4-bit MT | 0.266 | 3,758,492 | 0.228 |
| d=3072 2-bit ST | 0.213 | 4,695,736 | 0.177 |
| d=3072 2-bit MT | 0.257 | 3,891,492 | 0.209 |
| d=3072 4-bit ST | 0.303 | 3,298,099 | 0.230 |
| d=3072 4-bit MT | 0.302 | 3,314,459 | 0.261 |

Cross-checks on the recorded numbers:

- Bulk ST at d=1536 — 37.4K (2-bit) / 34.7K (4-bit) vec/s — sits 1.5-1.6× above #57's ad-hoc encode bench (25.7K / 21.9K on an M2 Pro): same order, faster host. Warm append (steady-state encode, calibration frozen) runs 3-4× above bulk, since bulk carries the calibration fit and one-time init inside the timed first `add()`.
- MT/ST bulk scaling is 3.6-4.0× across all four (dim, bit) cells — the Rayon encode parallelism from #57 is what these cells will regression-guard. Warm append scales much less under MT (1.1-1.9×): at these batch sizes it is dominated by the rotation matmul, which is BLAS-threaded even in ST cells (`RAYON_NUM_THREADS` doesn't pin Accelerate, same as the existing suite).
- Removal is O(1)-flat: 0.17-0.30µs per op everywhere (d=3072 modestly above d=1536, 4-bit above 2-bit, as code size predicts), with the id-map layer adding ~0.02-0.07µs of bookkeeping over raw `swap_remove` and MT cells reading ~1.1-1.3× ST (removal is serial; the gap tracks Rayon worker spin-down after the untimed MT index build).
- FAISS bulk add is faster than TQ bulk in every cell — expected: `IndexPQFastScan.add()` encodes against already-trained codebooks with no per-vector d×d rotation. Recorded as measured; no prose claims drawn from it.

Measurement notes: all 16 cells were measured sequentially in a single continuous window on an otherwise-idle machine (foreground shell, the same way `README → Running benchmarks` documents running the suite). An earlier version of these results — pushed in this branch's first commit — was measured while unrelated compute ran on the host and read 1.5-3× pessimistic; those JSONs are replaced by this window's. Cross-checking runs from different windows: TQ bulk/warm and the remove medians reproduce within ~6%, while `tq_single_add_us` and the FAISS baseline are the most sensitive to host scheduling state.

## x86 results (c3-standard-8, Intel Sapphire Rapids — same host as the published x86 search cells)

Measured on a Google Cloud `c3-standard-8` (Intel Sapphire Rapids, 8 vCPUs), release build, median of 5 runs, otherwise-idle machine — same methodology and cell matrix as the ARM results above.

**Insertion**

| Cell | TQ bulk (vec/s) | TQ warm append (vec/s) | TQ single add (µs/op) | FAISS bulk (vec/s) |
|---|---:|---:|---:|---:|
| d=1536 2-bit ST | 10,561 | 22,868 | 256.5 | 26,546 |
| d=1536 2-bit MT | 32,090 | 42,429 | 257.7 | 147,821 |
| d=1536 4-bit ST | 8,526 | 15,920 | 283.1 | 31,775 |
| d=1536 4-bit MT | 28,180 | 39,143 | 282.4 | 120,667 |
| d=3072 2-bit ST | 4,268 | 10,105 | 1185.2 | 13,894 |
| d=3072 2-bit MT | 11,616 | 17,266 | 1161.7 | 76,404 |
| d=3072 4-bit ST | 3,606 | 7,039 | 1229.5 | 15,713 |
| d=3072 4-bit MT | 10,602 | 15,325 | 1232.8 | 60,218 |

**Removal**

| Cell | IdMap remove (µs/op) | IdMap remove (ops/s) | swap_remove (µs/op) |
|---|---:|---:|---:|
| d=1536 2-bit ST | 0.414 | 2,414,925 | 0.352 |
| d=1536 2-bit MT | 0.417 | 2,398,418 | 0.358 |
| d=1536 4-bit ST | 0.622 | 1,606,659 | 0.464 |
| d=1536 4-bit MT | 0.618 | 1,618,045 | 0.474 |
| d=3072 2-bit ST | 0.624 | 1,603,434 | 0.478 |
| d=3072 2-bit MT | 0.599 | 1,669,999 | 0.468 |
| d=3072 4-bit ST | 0.968 | 1,033,440 | 0.759 |
| d=3072 4-bit MT | 0.983 | 1,017,356 | 0.741 |

Reading the numbers: MT/ST bulk-insert scaling runs ~2.7-3.3× across the four (dim, bit) cells; warm append scales ~1.7-2.5× (matmul-bound, as on ARM). Removal is O(1)-flat at 0.41-0.98µs per op, the id-map layer adding ~0.06-0.24µs over raw `swap_remove`, with no material ST/MT gap. FAISS bulk add is faster than TQ bulk in every cell, as on ARM (codebooks pre-trained, no per-vector rotation). Recorded as measured; no prose claims drawn from it.

## README figures (maintainer ruling)

Per the maintainer's ruling during review (supersedes the earlier "results land, README untouched" framing in this PR's first version, and #65's original "prose mention is enough / diagram out of scope" note): the README gains an **Insertion & Removal Speed** section directly after Search Speed, with per-architecture figures generated by `create_diagrams.py` from the committed JSONs, in the existing figure style:

- `docs/arm_insert_st.svg`, `docs/arm_insert_mt.svg`, `docs/x86_insert_st.svg`, `docs/x86_insert_mt.svg` — TQ bulk / TQ warm append / FAISS bulk, per (dim, bit) cell
- `docs/arm_remove_st.svg`, `docs/x86_remove_st.svg` — `IdMapIndex.remove` vs `swap_remove` per-op latency, per (dim, bit) cell

The README section states the measurement setup and links the suite and result files; no claims beyond what the JSONs show. Both ARM and x86 figures are now present (ARM and x86 subsections mirror the Search Speed layout). Existing SVGs regenerate byte-identical.

## Notes

- CHANGELOG: entries under `[Unreleased] → Benchmarks` and `→ Docs`.
- No CI changes.
- Full `cargo test --workspace` (208 passed / 0 failed) and `pytest` (552 passed) run clean; the suite scripts are additive and nothing under test changed.


skulmakov-oss:
Hi @RyanCodrai,

This benchmark split between **bulk initialization** and **warm steady-state
operations** is particularly useful.

It captures something we also observed while integrating TurboVec into
Semantic Hub: there are really two different performance problems here, and
they benefit from two complementary solutions.

At the integration layer, our approach is to keep TurboVec behind a
long-lived, supervised Hub worker:

```text
CLI / application
    -> Semantic Hub host
    -> persistent TurboVec worker
    -> warm in-memory index

The index is loaded once, then subsequent insert, search, filtered-search, and
remove operations reuse the already-initialized rotation, codebook, calibration,
and search layout.

That turns the approximately 250–300 ms cold reconstruction cost we measured
into an initialization cost rather than a per-request cost. On our small
qualification workload, warm search was approximately 0.085 ms, while the Hub
admission, resource-budget, panic-boundary, audit, and provenance layer added
only approximately 125–150 µs.

The second layer is the upstream cache/sidecar direction discussed elsewhere:
even a long-lived worker must occasionally start cold after a machine restart,
worker restart, version transition, or memory eviction. Persisting validated
derived state makes that unavoidable cold start substantially cheaper.

So the combined architecture becomes:

normal operation:
    persistent Hub worker
    -> warm TurboVec index
    -> no repeated reconstruction

real cold start:
    .tvim + validated backend-specific sidecar
    -> fast worker restoration
    -> warm operation resumes

Your decision in this PR to measure bulk initialization separately from warm
append is exactly the right benchmark structure for evaluating both layers.
A single aggregate number would hide where the cost actually belongs.

We are applying this model in the Semantic Hub integration here:

https://github.com/skulmakov-oss/Semantic/pull/1554

Project:

https://github.com/skulmakov-oss/Semantic

One reason TurboVec was such a good first real Hub tool is that its design
exposes these phases clearly: index construction, calibration, persisted state,
derived runtime state, warm mutation, and search are distinguishable rather
than being hidden behind one opaque service boundary.

That clarity makes it possible for downstream systems to solve lifecycle
amortization correctly while upstream TurboVec continues improving the genuine
cold-start path.

Very strong benchmark design.

skulmakov-oss:
## Independent Windows/x86 qualification

**Environment**

- Windows 11 Pro Insider Preview 10.0.26300; Intel Core i5-9300H (4 physical / 8 logical cores); 19.87 GiB RAM
- Python 3.13.14; Rust/Cargo 1.97.0; NumPy 2.5.1 (OpenBLAS 0.3.33.112.0); FAISS 1.14.3 (VCOMP/OpenBLAS)
- The canonical benchmark datasets were not available locally, and this host is below the reference x86 qualification profile. I ran correctness-only synthetic ST/MT smoke checks; I did **not** create or publish canonical x86 numbers or x86 result JSON.

**Reviewed exact head**

- `95d2892ded6e8449cfbb55756759abc1322ec060`; TurboVec wheel built locally from this commit

**Findings**

- **CONFIRMED:** the insert scripts took five single-add samples from one progressively growing index, so the samples were not measured from the same post-warm state. The local patch measures each sample from a fresh 110K-vector state. If accepted, the existing ARM `tq_single_add_us` values need remeasurement by Ryan; recorded ARM JSON was not changed.
- **CONFIRMED:** `create_diagrams.py` used the locale-default output encoding and direct writes. On this CP1251 Windows host it raised `UnicodeEncodeError` on the multiplication sign after truncating `docs/compression.svg`. The patch uses one UTF-8 atomic writer for all SVGs.
- **CONFIRMED:** benchmark result JSON used direct final-path writes and scripts did not reject a mismatched architecture. The patch adds a small shared helper for temp-file replacement and an explicit ARM/x86 guard.
- **DOCUMENTATION AMBIGUITY / MEASUREMENT LIMITATION:** ST pins Rayon and FAISS OpenMP to one thread, but does not control the Linux/macOS BLAS backend used by TurboVec rotation. The patch defines ST precisely and labels the affected ARM figures “Rayon single-threaded”; metric values remain untouched.
- **NOT REPRODUCED:** matrix/schema/filename/key/seed/FAISS-layout drift. All 48 scripts form six internally consistent templates; all 16 ARM result files pass schema checks, and remove throughput agrees with latency within JSON rounding. Raw and `IdMap` removal schedules differ in exact order but are equivalent uniform workloads.
- **MEASUREMENT LIMITATION:** the committed JSON keeps medians but not per-run variance or host-load metadata. I did not expand the format because that would be a broader methodology change.

**Validation**

- `cargo test --workspace`: pass
- `cargo test -p turbovec --release`: 208 tests + 3 doctests pass
- downstream smoke: pass (search plus write/load round-trip)
- benchmark helper tests: 3 pass; all 48 scripts compile and pass the matrix/AST audit; diagram generation is byte-stable; README links and all 16 ARM JSON schema/math checks pass; `git diff --check` passes
- Python suite: 551 pass, 1 fail. The failure is the untouched `test_delete_vs_delete[agno]` duration-based concurrency case; two runs ended without errors or map desynchronization but before deleting the full set. Current unpinned Agno resolved to 2.8.3. I treated this as out of scope rather than changing storage code.
- `cargo fmt --all --check`: fails only on pre-existing, untouched Rust formatting under rustfmt 1.97.0; no Rust files were changed.
- Current CI on the reviewed PR head: 6/6 pass

The cold-versus-warm separation in this benchmark design is the correct abstraction for real downstream runtimes such as Semantic Hub: https://github.com/skulmakov-oss/Semantic/pull/1554

I prepared local commit `d2c7ce3cef571fc07b7473c4591a281fcb4f0e3c` based directly on the reviewed head. It is not pushed. Ryan, please let me know whether you prefer a patch, a helper PR targeting `feat/insert-remove-benchmarks`, or direct commits to that branch.

